PR #9, opened 2026-06-10
Added listener-side support bits for QUIC connection migration

mishtak00:
- Declared new `hasMigrationInfo` flag on `PathProperties` and wired it in `ManyToManyProtocol`
- Propagated new path event to QUIC via `attachLowerDatagramProtocolForNewPath`

### DIFF
--- a/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
@@ -246,6 +246,7 @@ public protocol MultiplexingPath: UpperProtocolHandler {
     var identifier: MultiplexingPathIdentifier { get }
     init(parent: ParentProtocol)
     var pathIsPrimary: Bool { get set }
+    var pathHasMigrationInfo: Bool { get set }
 }
 
 @_spi(ProtocolProvider)
@@ -969,8 +970,13 @@ extension ManyToManyDatapathProtocol where Path.ParentProtocol == Self, Path: In
             parameters: parameters,
             path: path
         )
-        if multiplexingPaths.isEmpty { newPath.pathIsPrimary = true }
+        let isFirstPath = multiplexingPaths.isEmpty
+        if isFirstPath { newPath.pathIsPrimary = true }
+        if path?.hasMigrationInfo == true { newPath.pathHasMigrationInfo = true }
         multiplexingPaths[newPath.identifier] = newPath
+        if !isFirstPath {
+            handlePathChanged(path: newPath.identifier, event: .available, isPrimary: newPath.pathIsPrimary)
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


agnosticdev:
It looks like `testQUICSpinBitEnabledOrDisabled` failed.  Kicked off the tests again to see if this was an intermittent failure.

agnosticdev (approved):

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -932,8 +933,13 @@ extension ManyToManyDatapathProtocol where Path.ParentProtocol == Self, Path: In
             parameters: parameters,
             path: path
         )
-        if multiplexingPaths.isEmpty { newPath.pathIsPrimary = true }
+        let isFirstPath = multiplexingPaths.isEmpty
+        if isFirstPath { newPath.pathIsPrimary = true }
+        if path?.hasMigrationInfo == true { newPath.pathHasMigrationInfo = true }
         multiplexingPaths[newPath.identifier] = newPath
+        if !isFirstPath {
+            handlePathChanged(path: newPath.identifier, event: .available, isPrimary: newPath.pathIsPrimary)
```
rpaulo, line 978: Why is the event `.available` ? Shouldn't it be `.established`?
Why are we only sending an event when it's not the first path?

mishtak00: As it stands, `.available` gets fed to `QUICConnection.handlePathChanged(.available)`, which calls `path.changeState(to: .routeAvailable)` and `invokeEstablish()`, which calls `connect()`. The `.established` event does `beginValidation()`, which is further down the line for migration. Are you suggesting we wiggle this state wiring a bit?

As for why not first path too, `initialPath` has a different treatment. See QUICConnection.swift:1291. From my understanding, first path is a different concept from primary path. How could we even kick off migration if the target path is the first path? Or are you saying we'd be migrating back to the initial path from a sojourn?

rpaulo: Well, available just means that the system told us we can attach a lower protocol to it.  Here you are attaching a lower protocol already, so I thought it would be .established.

mishtak00: > Here you are attaching a lower protocol already, so I thought it would be .established.

The call chain is:
protocol stack linkage (attachLowerProtocolForNewPath etc) -> .available path event -> invokeEstablish -> invokeConnect -> connect propagates down the stack -> .connected propagates up -> .established path event.

We might need @tfpauly to comment here too.

rpaulo: Discussed with @tfpauly, this will change in the future but lets go with this change for now.

rpaulo (approved):

PR #14, opened 2026-06-12
agnostivdev/StreamMetadata: Make QUICStreamMetadata a non-copyable struct

agnosticdev:
This is a performance change to reduce the amount of reference types allocated when creating a high volume of QUIC streams.  This will be particularly noticeable on the server side when QUIC is servicing a high volume of streams.

I did some other independent CPU analysis and this change also reduces the CPU load when dealing with high volume stream use cases.


agnosticdev:
This one needs to wait until Swift 6.4 is more broadly supported and we can adopt it in our CI due to
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0503-suppressed-associated-types.md

PR #15, opened 2026-06-12
agnosticdev/RemoveLazyVars: Remove lazy vars in ProtocolStack

agnosticdev:
Remove the use of `lazy` vars in the ProtocolStack to make sure protocols are always available in-memory when a consuming application uses them.


agnosticdev:
Fixed in https://github.com/apple/swift-network-evolution/pull/21

PR #16, opened 2026-06-13
Fix QUICTransfer

agnosticdev:
In the current pattern the QUICTransfer benchmark did not work due to `ProtocolStreamHandlers` having an empty `upperReceiveQueue`. Now the reads are scheduled on a new async block to give the original write time to unwind and process through the stack.

I double checked `IPUDPTransfer`, `QUICHandshake`, and `QUICStreamLoad` also and they seem to be fine.

### DIFF
--- a/Sources/Tools/QUICTransfer/main.swift
+++ b/Sources/Tools/QUICTransfer/main.swift
@@ -43,11 +43,13 @@ final class QUICTransfer {
     let NSEC_PER_MSEC = UInt64(Duration.milliseconds(1) / Duration.nanoseconds(1))
     var serverSigningKey = P256.Signing.PrivateKey()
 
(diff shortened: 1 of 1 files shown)


tfpauly:
I think the more correct solution here is to model off of what QUICStreamLoad does and use `BridgeDatagramProtocol`.

agnosticdev:
> I think the more correct solution here is to model off of what QUICStreamLoad does and use `BridgeDatagramProtocol`.

Yeah, we can deprecate the old pattern and migrate to `BridgeDatagramProtocol`.  IPUDPTransfer still needs to adopt `BridgeDatagramProtocol` as well so I will open an issue there.

tfpauly (approved):

Issue #19, opened 2026-06-16
Support for static Linux.


zaneenders:
Hi 👋,

When building the [swift-nio-quic](https://github.com/apple/swift-nio-quic) and got an error for the static Linux SDK.

```
➜  swift-network-evolution git:(main) swift build -c release --swift-sdk x86_64-swift-linux-musl

[1/1] Planning build
Building for production...
In file included from /home/zane/Developer/swift-network-evolution/Sources/SwiftNetworkLinuxShim/SwiftNetworkLinuxShim.c:21:
/home/zane/Developer/swift-network-evolution/Sources/SwiftNetworkLinuxShim/include/SwiftNetworkLinuxShim.h:21:10: fatal error: 'linux/netlink.h' file not found
   21 | #include <linux/netlink.h>
      |          ^~~~~~~~~~~~~~~~~
1 error generated.
[1/11] Compiling SwiftNetworkLinuxShim SwiftNetworkLinuxShim.c
```

Setup 

```
➜  swift-network-evolution git:(main) swift --version
Swift version 6.3.2 (swift-6.3.2-RELEASE)
Target: x86_64-unknown-linux-gnu

➜  swift-network-evolution git:(main) uname -a
Linux tower 6.8.0-124-generic #124-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 13:00:45 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

➜  swift-network-evolution git:(main)   cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.4 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo
```

Also got the error on MacOS(Version 26.5.1 (25F80)

Thanks,
Zane

rnro:
[...]

PR #23, opened 2026-06-17
Disable Netlink to unblock swift-nio-quic builds

agnosticdev:
This PR disables the Netlink functionality until we devise a strategy to support this functionality long term on all Linux variants.  This will unblock the swift-nio-quic builds and hopefully address this issue: https://github.com/apple/swift-network-evolution/issues/19

### DIFF
--- a/Sources/SwiftNetwork/Endpoint/NetlinkAddress.swift
+++ b/Sources/SwiftNetwork/Endpoint/NetlinkAddress.swift
@@ -35,7 +35,7 @@ struct NetlinkAddress: IPAddress, Hashable, CustomDebugStringConvertible {
     /// Creates a Netlink address from raw bytes.
     init?(_ bytes: [UInt8]) {
         if bytes.count >= NetlinkAddress.layoutSize {
-            #if os(Linux)
+            #if os(Linux) && NETLINK_ENABLED
             let nl = bytes.withUnsafeBytes { $0.load(as: sockaddr_nl.self) }
             self.init(nl)
             #else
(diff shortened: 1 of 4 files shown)


tfpauly (approved):

Sources/SwiftNetwork/Endpoint/NetlinkAddress.swift
```diff
@@ -35,7 +35,7 @@ struct NetlinkAddress: IPAddress, Hashable, CustomDebugStringConvertible {
     /// Creates a Netlink address from raw bytes.
     init?(_ bytes: [UInt8]) {
         if bytes.count >= NetlinkAddress.layoutSize {
-            #if os(Linux)
+            #if os(Linux) && NETLINK_ENABLED
```
rnro, line 38: Isn't this and other changes actually _widening_ the enablement of this code path rather than disabling it?

josephnoir:
What is the path to enable this as an adopter, i.e., does it need a trait?

Sources/SwiftNetwork/Endpoint/NetlinkAddress.swift
agnosticdev: The thought here was that enabling this flag would allow us to selectively re-enable this when adding it back in the future.

rnro: Sorry, for some reason I read the diff in reverse🙃. Makes sense.

agnosticdev:
> What is the path to enable this as an adopter, i.e., does it need a trait?

As an adopter if you wanted to adopt or work on this you would need add in `NETLINK_ENABLED` as a flag.  At this point I think we should do some more work on this part of the stack to modernize it a bit and add some forward declarations to properly expose Netlink in all Linux variants.  Since this path is not actively used I do not think we need to add a trait right now.

josephnoir (approved):

rnro (approved):
I am not sure about some of the details here but I think this serves the main purpose of unblocking adopters and we can iterate

PR #26, opened 2026-06-18
CI improvements

rnro:
- Cover every Darwin platform via xcodebuild (macOS, Mac Catalyst, iOS, watchOS, tvOS, visionOS) alongside the existing `xcrun swift test`. Every step runs regardless of earlier failures and a final summary table reports per-platform `PASSED`/`FAILED`.
- On Linux, build the package with three trait combinations before running the test suite: 
  - default
  - all additive traits on (`DatapathLogging` / `QlogOutput` / `SignpostOutput`)
  - all reductive traits on (`DisableDebugLogging` / `DisableErrorLogging`).
- Tests run in both debug and release.
- Enable the Static Linux Swift SDK build job.
- Bump the `swiftlang/github-workflows` pin from `0.0.9` to `0.0.11` to pick up updated static-SDK defaults (the `0.0.9` default matrix installs Swift 6.2 toolchains, which can't satisfy our 6.3 `swift-tools-version`).
- Add a Dependabot config that groups `swiftlang/github-workflows/*` bumps so future pin updates arrive as a single PR rather than one per workflow file.
- Clarify workflow results.
  - Pass `--quiet` to every `swift build` / `swift test` and to `xcrun swift test`
  - emit `=== <step>: PASSED ===` markers per step (plus a column-aligned summary on macOS and an "all builds and tests passed" footer on Linux)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      swiftlang-actions:
+        patterns:
+          - "swiftlang/github-workflows/*"
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   tests:
     name: Test
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.9
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.11
     with:
       swift_flags: "-Xswiftc -warnings-as-errors"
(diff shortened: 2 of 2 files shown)


rnro:
Static SDK failures are expected due to a `Glibc` import issue which is being tackled separately.

rpaulo (approved):

agnosticdev (commented):
Thank you very much for adding this CI support! I think this will be important to match the similar build jobs that swift-nio-quic has. Does this also build for watchOS arm64_32?

rnro:
> Thank you very much for adding this CI support! I think this will be important to match the similar build jobs that swift-nio-quic has. Does this also build for watchOS arm64_32?

This builds in exactly the same ways as swift-nio-quic (same commands), just within the swiftlang CI paradigm.

agnosticdev (approved):

PR #29, opened 2026-06-18
Track pending wakeup state in Timer

glbrntt:
Timer's optimisation in 'recalculate' for an already scheduled wakeup applies even if there's no wakeup pending. This can happen when the wakeup is early as an event which would've fallen within the threshold no longer does, so it can remain unscheduled.

Store the kernel wakeup state as an enum: it's either armed for a given time or idle.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/Timer.swift
+++ b/Sources/SwiftNetwork/QUIC/Timer.swift
@@ -86,9 +86,9 @@ final class Timer: PrefixedLoggable {
             return .zero
         }
     }
-    private var wakeup: KernelWakeup = .idle
+    private var wakeup: WakeupState = .idle
 
-    private enum KernelWakeup {
+    private enum WakeupState {
         case idle
         case armed(NetworkClock.Instant)
     }
--- a/Sources/SwiftNetwork/QUIC/Timer.swift
+++ b/Sources/SwiftNetwork/QUIC/Timer.swift
@@ -78,7 +78,20 @@ final class Timer: PrefixedLoggable {
     #else
     private let extraDebugging = false
     #endif
-    private(set) var nextDeadline: NetworkClock.Instant = .zero
+    var nextDeadline: NetworkClock.Instant {
+        switch wakeup {
+        case .armed(let instant):
+            return instant
+        case .idle:
+            return .zero
+        }
+    }
+    private var wakeup: KernelWakeup = .idle
+
+    private enum KernelWakeup {
+        case idle
+        case armed(NetworkClock.Instant)
+    }
 
     static let timerThreshold = NetworkDuration.milliseconds(1)
 
@@ -126,6 +139,7 @@ final class Timer: PrefixedLoggable {
         if !timerCancelled {
             log.debug("Stopping timer")
             timerCancelled = true
+            wakeup = .idle
             reference?.unscheduleWakeup()
         }
         if final {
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/QUIC/Timer.swift
```diff
@@ -78,7 +78,20 @@ final class Timer: PrefixedLoggable {
...
-    private(set) var nextDeadline: NetworkClock.Instant = .zero
+    var nextDeadline: NetworkClock.Instant {
+        switch wakeup {
+        case .armed(let instant):
+            return instant
+        case .idle:
+            return .zero
+        }
+    }
+    private var wakeup: KernelWakeup = .idle
```
tfpauly: Does it make sense to call this "kernel" always? Maybe some other term to indicate it is the actual lower state?

Sources/SwiftNetwork/QUIC/Timer.swift
agnosticdev, line 1: Do running these timer values through an enum cause a CPU issue that outweighs the gain?  For example if you try this with QUICTransfer top of tree and with this change do you see a difference either way?

glbrntt: No, I did five runs of each:

```
with this patch:  24.53s / 24.92s / 24.65s   (min / max / mean)
        without:  24.85s / 25.06s / 24.95s   (min / max / mean)
```

With this patch is _marginally_ faster but I expect that if anything it's down to noise.

Sources/SwiftNetwork/QUIC/Timer.swift
glbrntt: Good point, renamed to `WakeupState`

agnosticdev (approved):

Sources/SwiftNetwork/QUIC/Timer.swift
```diff
@@ -126,6 +139,7 @@ final class Timer: PrefixedLoggable {
         if !timerCancelled {
             log.debug("Stopping timer")
             timerCancelled = true
```
rpaulo, line 141: do we still need `timerCancelled` after these changes?

Tests/QUICTests/TimerTests.swift
```diff
@@ -179,6 +179,82 @@ final class TimerTests: XCTestCase {
...
+        // B, 999us after A (crucially, just before the 1ms coalescing threshold)
+        let idB = timer.insert(
+            description: "B",
+            fromNow: .seconds(2) + .microseconds(999),
+            timerNow: .zero
+        ) {
+            semaphore.signal()
+        }
+        // A was earlier, deadline is unchanged
+        XCTAssertEqual(timer.nextDeadline, .init(.seconds(2)))
```
rpaulo, line 202: Do we really need `.init()` here?

rpaulo (approved):

Tests/QUICTests/TimerTests.swift
glbrntt: Yes: `nextDeadline` is an instant and `.seconds` is a duration, `.init(.seconds(2))` gives us an instant 2 seconds after t=0

Sources/SwiftNetwork/QUIC/Timer.swift
glbrntt: Maybe, maybe not. There are some separate branches on `timerCancelled` but it's not clear if they're actually necessary. 

I didn't want to scope creep this change from a targeted fix for the observed bug to refactoring the internal state management of the timer. That may very well be a worthwhile change though.

PR #30, opened 2026-06-19
Add support for Musl

josephnoir:
Static SDK builds on Linux use Musl instead of Glibc. This adds import guards to switch to Musl when available. Two other changes are necessary to adopt to differences between the two (check `SystemSocket.swift`):

* `SOCK_{STREAM,DGRAM,RAW}` do not have a `raw_value` member in Musl
* `sin6_addr` members have different names

Building with the netlink flag revealed a problem with the imports there. Adjusting the guard around importing the headers resolves that.

### DIFF
--- a/Sources/SwiftNetwork/System/SystemAddressFamily.swift
+++ b/Sources/SwiftNetwork/System/SystemAddressFamily.swift
@@ -13,7 +13,11 @@
 //===----------------------------------------------------------------------===//
 
 #if os(Linux)
+#if canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
+#endif
 #elseif !NETWORK_STANDALONE
 import Darwin
 #endif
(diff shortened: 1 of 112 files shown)


rnro (approved):
Looks good, thank you!

Sources/SwiftNetwork/System/SystemAddressFamily.swift
```diff
@@ -13,7 +13,11 @@
 //===----------------------------------------------------------------------===//
 
 #if os(Linux)
```
rnro: Is this outer check still needed or can it be a strait Glibc/Musl/not standalone check? I don't really have a strong opinion which is better even if it is technically valid to have the structure flat, only a weak preference.

josephnoir: This can indeed be flattened. I don't have a strong preference either. Removing lines makes it a bit easier to parse, so addressed in 28a6ab2.

Sources/SwiftNetworkLinuxShim/include/SwiftNetworkLinuxShim.h
agnosticdev, line 1: Thank you

agnosticdev (approved):

rpaulo (approved):

PR #32, opened 2026-06-27
Add reassembly support for fragmented IPv4 packets in Swift IP

agnosticdev:
The goal here is to add reassembly support for fragmented IPv4 packets in Swift IP without impacting performance and on standard path.
Since IP fragmentation is not very common there needs to be fall-back behavior for reassembly when IP fragments are detected in `processInboundFrames` to ensure performance.

This change includes such fall-back behavior for reassembly and I verified the performance with QUICTransfer:
```
Top of tree:
213.05 M  47.9%	-	 IPProtocol.IPInstance.IPv4Instance.processInboundFrames(_:_:)

With this change
226.35 M  46.7%	-	 IPProtocol.IPInstance.IPv4Instance.processInboundFrames(_:_:)
```

Adds support for reassembly IPv4 fragments only. IPv6 support will come in a future change.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -121,7 +121,8 @@ let package = Package(
             name: "SignpostOutput",
             description: "Enables `OSSignposter` output from the QUIC implementation."
         ),
-        .default(enabledTraits: []),
+        //.default(enabledTraits: []),
+        .default(enabledTraits: ["DisableDebugLogging", "DisableErrorLogging"]),
     ],
     dependencies: packageDependencies,
     targets: [
--- a/Sources/SwiftNetwork/Protocols/FrameArray.swift
+++ b/Sources/SwiftNetwork/Protocols/FrameArray.swift
@@ -57,6 +57,19 @@ public struct FrameArray: ~Copyable {
         }
     }
 
+    public mutating func prepend(frame: consuming Frame) {
+        var newArray = FrameArray(capacity: self.count + 1)
+        newArray.add(frame: frame)
+        if !self.frames.isEmpty {
+            while !self.frames.isEmpty {
+                if let first = self.frames.popFirst() {
+                    newArray.add(frame: first)
+                }
+            }
+        }
+        self = newArray
+    }
+
     public var count: Int {
         frames.count
     }
--- a/Sources/SwiftNetwork/Protocols/FrameArray.swift
+++ b/Sources/SwiftNetwork/Protocols/FrameArray.swift
@@ -58,16 +58,7 @@ public struct FrameArray: ~Copyable {
     }
 
     public mutating func prepend(frame: consuming Frame) {
-        var newArray = FrameArray(capacity: self.count + 1)
-        newArray.add(frame: frame)
-        if !self.frames.isEmpty {
-            while !self.frames.isEmpty {
-                if let first = self.frames.popFirst() {
-                    newArray.add(frame: first)
-                }
-            }
-        }
-        self = newArray
+        self.frames.prepend(frame)
     }
 
     public var count: Int {
--- a/Sources/SwiftNetwork/Protocols/IPProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/IPProtocol.swift
@@ -339,6 +339,10 @@ public struct IPProtocol: NetworkProtocol {
         var log = NetworkLoggerState()
         var eventManager = ProtocolEventManager()
 
+        static let IP_MF: UInt16 = 0x2000
+        static let IP_OFFMASK: UInt16 = 0x1FFF
+        static let IP_MAX_FRAGMENT_COUNT: Int = 32
+
         // Only called by newProtocolInstance()
         fileprivate static func registerNewIP(on context: NetworkContext) -> ProtocolInstanceReference {
             let ip = IPInstance(context: context)
@@ -352,7 +356,7 @@ public struct IPProtocol: NetworkProtocol {
 
         var passthroughEvents = true
 
-        struct IPCounters {
+        struct IPCounters: ~Copyable {
             var txPackets = 0
             var rxPackets = 0
             var rxECT0Packets = 0
@@ -367,9 +371,9 @@ public struct IPProtocol: NetworkProtocol {
             var dscpValue: UInt8?
         }
 
-        struct IPReassemblyState {
-            var reassemblyID: UInt32
-            //	TODO: inputReassemblyFrames, assignedInputFrameArray, assignedOutputFrameArray
+        struct IPReassemblyState: ~Copyable {
+            var reassemblyID: UInt16
+            var inputReassemblyFrames = FrameArray()
         }
 
         struct IPInstanceFlags: OptionSet {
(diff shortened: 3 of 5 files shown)


agnosticdev:
Outbound fragmentation support for IPv4 will also come in a future change.

Sources/SwiftNetwork/Protocols/FrameArray.swift
```diff
@@ -57,6 +57,19 @@ public struct FrameArray: ~Copyable {
         }
     }
 
+    public mutating func prepend(frame: consuming Frame) {
```
tfpauly, line 60: The body of this function should just be
```
        self.frames.prepend(frame)
```

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -339,6 +339,10 @@ public struct IPProtocol: NetworkProtocol {
         var log = NetworkLoggerState()
         var eventManager = ProtocolEventManager()
 
+        static let IP_MF: UInt16 = 0x2000
```
tfpauly: Let's make these use more swift-styled names: `IPMoreFragmentsFlag`, etc.

Sources/SwiftNetwork/Protocols/FrameArray.swift
agnosticdev: Nice, that is much better! Addressed in[e8166d2](https://github.com/apple/swift-network-evolution/pull/32/commits/e8166d26d0cfe3afbb88ac46346665f59211b783)

Sources/SwiftNetwork/Protocols/IPProtocol.swift
agnosticdev: Addressed in [e8166d2](https://github.com/apple/swift-network-evolution/pull/32/commits/e8166d26d0cfe3afbb88ac46346665f59211b783)

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1026,12 +1321,37 @@ public struct IPProtocol: NetworkProtocol {
...
+                instance.processInboundFrames(log, &frames)
+                instanceType = .ipv4(instance)
+            case .ipv6(var instance):
+                instance.processInboundFrames(log, &frames)
+                instanceType = .ipv6(instance)
+            }
+        }
+
+        @inline(__always)
+        private static func processSend(_ instanceType: inout IPInstanceType, datagrams: inout FrameArray) {
```
tfpauly: To be parallel to processInbound, this should probably be processOutbound

agnosticdev: Addressed in [0b55f82](https://github.com/apple/swift-network-evolution/pull/32/commits/0b55f821a5eb336ff2c7376e05534719d693ae86)

tfpauly (approved):

Tests/SwiftNetworkTests/SwiftNetworkIPTests.swift
```diff
@@ -107,6 +107,52 @@ final class SwiftNetworkIPTests: NetTestCase {
     static let outputMessage: [UInt8] = [0xfc, 0xf7, 0x04, 0xd2, 0x00, 0x0d, 0xe8, 0x04, 0x68, 0x65, 0x6c, 0x6c, 0x6f]
     static let inputMessage: [UInt8] = [0x04, 0xd2, 0xfc, 0xf7, 0x00, 0x0d, 0xe8, 0x04, 0x68, 0x65, 0x6c, 0x6c, 0x6f]
 
+    // 169.254.225.163 -> 169.254.156.146, fragment 1 of 2: first 8 bytes of inputMessage, MF=1 (More fragments)
```
rpaulo: Given that these are IPv4LL, there would never be fragmentation in between them because there are no routers in between them :-)

Package.swift
```diff
@@ -121,7 +121,8 @@ let package = Package(
             name: "SignpostOutput",
             description: "Enables `OSSignposter` output from the QUIC implementation."
         ),
-        .default(enabledTraits: []),
+        //.default(enabledTraits: []),
```
rpaulo, line 124: Is this on purpose?

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -457,20 +461,233 @@ public struct IPProtocol: NetworkProtocol {
                 return value + IPv4Instance.headerLength
             }
 
-            mutating func processInboundFrames(_ ip: borrowing IPInstance, _ inboundFrames: inout FrameArray) {
+            mutating func appendReassembledPackets(
+                _ log: borrowing NetworkLoggerState,
+                reassembled: inout FrameArray
+            ) {
+                guard let empty = reassemblyState?.inputReassemblyFrames.isEmpty, !empty else {
```
rpaulo, line 468: I suggest guard let reassemblyState = reassemblyState {} right at the beginning and then drop the ?

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -457,20 +461,233 @@ public struct IPProtocol: NetworkProtocol {
...
+                }
+                guard let reassemblyID = reassemblyState?.reassemblyID else {
+                    return
+                }
+                var complete = false
+                var expectedOffset: UInt16 = 0
+                var tos: UInt8 = 0
+                var ttl: UInt8 = 0
+                reassemblyState?.inputReassemblyFrames.iterateMutableFrames { frame in
+                    guard frame.bufferLength >= IPv4Instance.headerLength else {
```
rpaulo, line 479: Do we have a test for every single if/guard for reassembly? We should because it's often a source of bugs.

agnosticdev: Added some more tests in [9f656f5](https://github.com/apple/swift-network-evolution/pull/32/commits/9f656f557ca0ecc410512dfb1836cf5e10f63079)

Sources/SwiftNetwork/Protocols/IPProtocol.swift
agnosticdev: If we do this then `reassemblyState` will have been consumed and we'd need to reinstantiate it.  Keep the optional.

Package.swift
agnosticdev: Good catch, its an artifact of the performance tests.

agnosticdev: Addressed in [9f656f5](https://github.com/apple/swift-network-evolution/pull/32/commits/9f656f557ca0ecc410512dfb1836cf5e10f63079)

Tests/SwiftNetworkTests/SwiftNetworkIPTests.swift
agnosticdev: Ha! Good point, yeah this is just for test purposes.  Added a comment on this.

PR #33, opened 2026-06-29
Drop `#if os(Linux)` from the package manifest

rnro:
Manifest `#if os(...)` is evaluated against the host running SwiftPM rather than the target, so it produces the wrong dependency graph when cross-compiling (e.g. building for Linux from macOS).

* Express the per-platform dependencies and the `SHIM_CRYPTO_SPAN_APIS` define via `.when(platforms:)` conditions instead of the `#if` branch.
* Inline the `packageDependencies` and `targetDependencies` bindings into the `Package` initializer, removing the now-empty `var` scaffolding.
* Trim `SwiftNetworkBenchmarks` to the dependencies it actually imports.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -125,7 +125,7 @@ let package = Package(
             dependencies: [
                 "SwiftNetwork",
                 .product(name: "SwiftTLS", package: "swift-tls"),
-                .product(name: "Crypto", package: "swift-crypto", condition: .when(platforms: [.linux])),
+                .product(name: "Crypto", package: "swift-crypto"),
                 .product(name: "Logging", package: "swift-log", condition: .when(platforms: [.linux])),
             ],
             swiftSettings: availabilityMacros + settings
--- a/Package.swift
+++ b/Package.swift
@@ -34,58 +34,27 @@ let availabilityMacros: [SwiftSetting] = versionNumbers.flatMap { version in
     }
 }
 
-var packageDependencies = [PackageDescription.Package.Dependency]()
-var targetDependencies = [PackageDescription.Target.Dependency]()
+let allApplePlatforms: [Platform] = [
+    .driverKit, .iOS, .macCatalyst, .macOS, .tvOS, .visionOS, .watchOS,
+]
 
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -34,59 +34,28 @@ let availabilityMacros: [SwiftSetting] = versionNumbers.flatMap { version in
...
+let settings: [SwiftSetting] = [
     .define("IMPORT_SWIFTTLS"),
     .define("EXPORT_SWIFTTLS"),
     .define("IMPORT_CRYPTO"),
     .define("SWIFTTLS_CERTIFICATE_VERIFICATION"),
+    // To support back to macOS 26, provide a shim on top of crypto APIs
+    // that allows passing spans. This is a less performant path, so for
+    // performance-sensitive cases, remove this define and require at least
+    // macOS 27.
+    .define("SHIM_CRYPTO_SPAN_APIS", .when(platforms: allApplePlatforms)),
```
agnosticdev, line 52: That should work, thanks.

Package.swift
```diff
@@ -123,11 +92,24 @@ let package = Package(
         ),
         .default(enabledTraits: []),
     ],
-    dependencies: packageDependencies,
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
```
agnosticdev, line 96: I realize that this isn't really part of this PR, but since we are here we might as well update the version of swift-log for Linux.

agnosticdev (approved):
Just one comment, not blocking though.

Package.swift
rnro: I can do that, but I'm not sure what it buys us. We have a `from:` dependency and the most recent version is `1.14.0` which will be used by projects when they update their dependencies if that works for the package graph. 

I think normally versions are only updated when they are _requiring_ some API/feature which is introduced in a given version.

agnosticdev: Yep, `from:` should work for us here.

tfpauly (approved):

PR #34, opened 2026-06-29
Add NAT64 support in Utilities

kpodosin:
Add initial support for NAT64. New files:

**Utilities/NAT64.swift:** 
- NAT64PrefixLength
- NAT64Prefix
- IPv4Address extensions: canBeSynthesizedNAT64 - checks if the address can be embedded into NAT64 prefix, isBlocklistedForWellKnownNAT64Prefix - checks for private/shared addresses that are excluded from the well-known prefix
- IPv6Address extensions: synthesized(from:prefix:) - get IPv6 address from IPv4 address and NAT64 prefix , extractedIPv4(using:) - get IPv4 address from NAT64-synthesized IPv6 address
- PathProperties extension: nat64Prefixes - decode _nat64Prefixes into typed [NAT64Prefix] values

**Utilities/IPv6Prefix.swift**
- struct IPv6Prefix

**SwiftNetworkTests/SwiftNetworkNAT64Tests**

### DIFF
--- a/Sources/SwiftNetwork/Utilities/NAT64.swift
+++ b/Sources/SwiftNetwork/Utilities/NAT64.swift
@@ -1,0 +1,155 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// The six prefix lengths defined by RFC 6052 for embedding IPv4 addresses into IPv6.
+///
+/// Raw values represent the prefix length in **bytes** (not bits). Use `bitLength`
+/// to get the length in bits.
+@available(Network 0.1.0, *)
+public enum NAT64PrefixLength: UInt8, Sendable {
+    case prefixLength32 = 4
+    case prefixLength40 = 5
+    case prefixLength48 = 6
+    case prefixLength56 = 7
+    case prefixLength64 = 8
+    case prefixLength96 = 12
+
+    var bitLength: Int { Int(rawValue) * 8 }
+
+    // Byte positions in a 16-byte IPv6 buffer where each of the 4 IPv4 bytes is placed.
+    // Byte 8 is always skipped (reserved per RFC 6052 §2.2).
+    var ipv4ByteOffsets: [4 of Int] {
+        switch self {
+        case .prefixLength32: return [4, 5, 6, 7]
+        case .prefixLength40: return [5, 6, 7, 9]
+        case .prefixLength48: return [6, 7, 9, 10]
+        case .prefixLength56: return [7, 9, 10, 11]
+        case .prefixLength64: return [9, 10, 11, 12]
+        case .prefixLength96: return [12, 13, 14, 15]
+        }
+    }
+}
+
+/// A NAT64 prefix — a base IPv6 address and length that defines the block of IPv6 address
+/// space used to represent IPv4 destinations on a NAT64 network.
+///
+/// Trailing bytes past `length` are always zeroed so that two prefixes with the same
+/// significant bits compare equal regardless of what was in the trailing positions.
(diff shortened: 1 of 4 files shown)


Sources/SwiftNetwork/Utilities/IPv6Prefix.swift
```diff
@@ -0,0 +1,18 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+struct IPv6Prefix {
+    var prefixAddress: (UInt32, UInt32, UInt32, UInt32) = (0, 0, 0, 0)
```
tfpauly: Let's make this be:

```suggestion
    var prefixAddress: IPv6Address = .any
```

Sources/SwiftNetwork/Utilities/NAT64.swift
```diff
@@ -0,0 +1,172 @@
...
+    case prefixLength48 = 6
+    case prefixLength56 = 7
+    case prefixLength64 = 8
+    case prefixLength96 = 12
+
+    var bitLength: Int { Int(rawValue) * 8 }
+
+    // Byte positions in a 16-byte IPv6 buffer where each of the 4 IPv4 bytes is placed.
+    // Byte 8 is always skipped (reserved per RFC 6052 §2.2).
+    var ipv4ByteOffsets: (Int, Int, Int, Int) {
```
tfpauly: I wonder if this should be a RangeSet instead of a tuple — or maybe just a function to extract the IPv4 bytes from a span view into the IPv6Address

Sources/SwiftNetwork/Utilities/NAT64.swift
```diff
@@ -0,0 +1,172 @@
...
+        self.address = IPv6Address(bytes)!
+    }
+
+    var debugDescription: String {
+        "\(address.debugDescription)/\(length.bitLength)"
+    }
+}
+
+extension IPv4Address {
+    private var isInLoopbackRange: Bool {
```
tfpauly: Seems like this is very similar to or duplicative of isLoopback

kpodosin: I think `isLoopback` only covers 127.0.0.1, but we want to cover the whole 127.0.0.0/8 range here.

Sources/SwiftNetwork/Utilities/NAT64.swift
```diff
@@ -0,0 +1,172 @@
...
+        v6[o1] = v4[1]
+        v6[o2] = v4[2]
+        v6[o3] = v4[3]
+        // Prefix bytes overwrite v6[0..<length]; ipv4ByteOffsets always places IPv4 bytes
+        // at positions >= length, so the prefix copy never clobbers an IPv4 byte.
+        withUnsafeBytes(of: prefix.address.address) { src in
+            v6.withUnsafeMutableBytes { dst in
+                dst.copyBytes(from: src.prefix(Int(prefix.length.rawValue)))
+            }
+        }
```
agnosticdev, line 125: Could any of this code above or in the constructor of `NAT64Prefix` be done with an InlineArray?

agnosticdev: Using an InlineArray avoids a temporary allocation.

tfpauly:
Please run `swift format -rip .` for formatting!

Sources/SwiftNetwork/Utilities/NAT64.swift
```diff
@@ -0,0 +1,155 @@
...
+    )
+
+    /// Creates a NAT64 prefix with the given length and base address.
+    ///
+    /// Bytes in `address` past `length` are zeroed so that equality and hashing
+    /// consider only the significant prefix bytes.
+    public init(length: NAT64PrefixLength, address: IPv6Address) {
+        self.length = length
+        var bytes: [16 of UInt8] = .init(repeating: 0)
+        withUnsafeBytes(of: address.address) { src in
```
tfpauly, line 71: We should avoid using unsafe bytes / unsafe mutable bytes as much as possible here. Doesn't need to be necessarily fixed in this PR, but moving to span access would help (EthernetAddress already has span accessors, as one example)

tfpauly (approved):

PR #36, opened 2026-06-30
Mark Sendable for NetworkContext's external async calls

tfpauly:
Adding Sendable marking to the async block for NetworkContext that can be used from off-context. The async call within the stack to schedule a block on the context itself validates that it is on the right context already, and is now called "isolatedAsync" within the NetworkContext.

This also requires a lot of changes to mark unsafe sendability in the tests that use async heavily along with expectations to do step-by-step testing.

### DIFF
--- a/Sources/SwiftNetwork/Connection/Connection.swift
+++ b/Sources/SwiftNetwork/Connection/Connection.swift
@@ -969,7 +969,7 @@ public struct TCP: StreamProtocol {
 #if !NETWORK_EMBEDDED
 @_spi(Essentials)
 @available(Network 0.1.0, *)
-public struct DatagramBridge: DatagramProtocol {
+public struct DatagramBridge: DatagramProtocol, Sendable {
     public typealias ContentType = Void
 
     public let belowProtocol: Void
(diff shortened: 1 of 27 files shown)


Sources/SwiftNetwork/EndpointFlow/EndpointFlow.swift
```diff
@@ -32,7 +32,7 @@ internal import Synchronization
 #endif
 
 @available(Network 0.1.0, *)
-final class EndpointFlow: CustomDebugStringConvertible {
+final class EndpointFlow: CustomDebugStringConvertible, @unchecked Sendable {
```
glbrntt, line 35: In NIO we have a rule that any use of `@unchecked Sendable` must include the rationale for why they are `@unchecked`. It's been incredibly valuable. Our long term goal is to remove all uses of `@unchecked`. You may find it helpful to adopt a similar pattern here.

Using `@unchcked` is often a footgun: it's easy to add new properties which aren't `Sendable` but might, for example, require being on the context to access safely. In NIO we usually isolate any state like that to its own struct within the type and only mark that as `@unchecked` but include runtime checks (for debug builds) to make sure we aren't violating those contracts. It was actually one such check that led us to this PR.

Sources/SwiftNetwork/Connection/Connection.swift
```diff
@@ -2114,10 +2116,12 @@ extension NetworkChannel where ApplicationProtocol: StreamProtocol {
         completion: (@Sendable (Result<Void, NetworkError>) -> Void)? = nil
     ) {
         let endpointFlow = self.endpointFlow
+        nonisolated(unsafe) let unsafeBuffer = buffer
+        nonisolated(unsafe) let unsafeOwner = owner
```
agnosticdev, line 2120: Are we able to build the WriteRequest and then pass it into the block to avoid the `nonisolated(unsafe) `?

PR #39, opened 2026-07-02
Avoid allocating in FrameArray empty init

glbrntt:
`FrameArray()` sets the minimum capacity to one which allocates a backing storage. This is a foot gun which is already triggered on a number of code paths: an empty `FrameArray()` is initialized only to be assigned over shortly after, resulting in an avoidable temporary allocation. 

This pattern is avoidable but takes care. A simpler solution which removes the foot gun altogether is to not request capacity in `FrameArray()`. When capacity is known users should use `FrameArray(capacity:)` instead.

This change reduces allocations by ~7% and CPU cycles by ~5% in the QUIC benchmarks I ran.

### DIFF
--- a/Sources/SwiftNetwork/Protocols/FrameArray.swift
+++ b/Sources/SwiftNetwork/Protocols/FrameArray.swift
@@ -36,7 +36,7 @@ public struct FrameArray: ~Copyable {
     }
 
     public init() {
-        self.frames = NetworkUniqueDeque<Frame>(minimumCapacity: 1)
+        self.frames = NetworkUniqueDeque<Frame>()
     }
 
     public init(capacity: Int) {
--- a/Sources/SwiftNetwork/QUIC/QUICConnection.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICConnection.swift
@@ -3196,11 +3196,14 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
         else {
             return false
         }
-        var datagramBatch = FrameArray()
+
+        var datagramBatch: FrameArray
         if self.flowControlState.pendingOutboundBytesToSend > 0 && availableCongestionWindow > 0 {
             datagramBatch = buildOutboundFrameBatch(
                 availableCongestionWindow: (availableCongestionWindow - totalSendBytes)
             )
+        } else {
+            datagramBatch = FrameArray()
         }
 
         // Sending on the current path.


tfpauly (approved):

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -3196,11 +3196,14 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
...
         }
-        var datagramBatch = FrameArray()
+
+        var datagramBatch: FrameArray
         if self.flowControlState.pendingOutboundBytesToSend > 0 && availableCongestionWindow > 0 {
             datagramBatch = buildOutboundFrameBatch(
                 availableCongestionWindow: (availableCongestionWindow - totalSendBytes)
             )
+        } else {
+            datagramBatch = FrameArray()
```
agnosticdev, line 3206: Here we should make the FrameArray allocate 0 and that makes total sense.

Sources/SwiftNetwork/Protocols/FrameArray.swift
```diff
@@ -36,7 +36,7 @@ public struct FrameArray: ~Copyable {
     }
 
     public init() {
-        self.frames = NetworkUniqueDeque<Frame>(minimumCapacity: 1)
+        self.frames = NetworkUniqueDeque<Frame>()
```
agnosticdev, line 39: The vast majority of the cases we have one or more frames in a FrameArray. It can happen that we do end up with an empty FrameArray and we end up wasting that allocation, this is not the standard case though.  The part that concerns me here is that for the majority of the cases we would start with zero storage in the frame array and then need to reallocate the size anyway.  The part you called out in QUICConnection absolutely makes sense as a place where it should have a zero capacity if nothing is going into the FrameArray.
I would be interested to see what the performance impact is, specifically in CPU, with this change in QUICTransfer and in QUICStreamLoad (the two benchmarks in SwiftNetwork here).  If those show good results then I think this is fine.  You'll want to look specifically at the CPU being expended when FrameArray is being used.

glbrntt: I couldn't get reliable instruments runs for QUICTransfer – there were large blocks in the middle of the run with no samples captured despite the run completing in about the right amount of time. 

For QUICStreamLoad:
- without the patch: 18.3 giga cycles
- with this patch: 16.8 giga cycles

So a reduction of 1.5 giga cycles or approximately 8%, which tallies with the 7% reduction I noted in the PR description. Runtime improvement (streams/s) was similar.

---

> The part that concerns me here is that for the majority of the cases we would start with zero storage in the frame array and then need to reallocate the size anyway. 

Empty containers shouldn't allocate – various container types have a shared "empty storage" backing (Array, Deque, etc.), when you first append to the container that's when they'll allocate, so you only pay the cost when you actually need to.

> The part you called out in QUICConnection absolutely makes sense as a place where it should have a zero capacity if nothing is going into the FrameArray.

That was one place, anecdotally there must be three or four others on the request path as the first time I tested this was just the change in `QUICConnection` (i.e. `FrameArray()` still allocating) and it saved one alloc per request. When I changed the `FrameArray` init it saved another three or four allocs per request, suggesting that this pattern isn't isolated to this one place.

glbrntt: FWIW: if there are many cases where FrameArray only stores one element it may be worth looking at a storing a single element inline before falling back to the heap if necessary (i.e. like TinyArray).

agnosticdev: I did some investigation here and as suspected QUICTransfer does do a trade of CPU cost for reallocating places where it did not previously have to, but ultimately for places where we are adding frames to the array we do see a slight gain of a few hundred megacycles here.  Just a few examples of where I am referring to:
```
Top of tree:
323.18 M  51.2%	35.01 M	  specialized QUICStreamInstance.dequeueReassembledData(connection:)

With this change:
272.52 M  46.3%	37.19 M	  specialized QUICStreamInstance.dequeueReassembledData(connection:)


With this change:
140.59 M  20.1%	-	      specialized QUICConnection.buildSinglePacketForKeyState()
140.59 M  20.1%	     	    FrameArray.add(frame:)


Top of tree:
106.95 M  17.9%	-	      specialized QUICConnection.buildSinglePacketForKeyState()
106.95 M  17.9%	            FrameArray.add(frame:)
```

agnosticdev: In QUICStreamLoad I see in aggregate a slight win:
```
Top of tree:
11.32 G  100.0%	-	 QUICStreamLoad (71791)

With this change:
11.12 G  100.0%	-	 QUICStreamLoad (36593)
```
And the same as QUICTransfer, wins and losses but ultimately we come up ahead.

rpaulo (approved):

agnosticdev (approved):

PR #41, opened 2026-07-02
SwiftQUIC: Fix variance in Spin Bit testing

agnosticdev:
Lately I have been noticing that the test for `testQUICSpinBitEnabledOrDisabled` has become increasingly unreliable. This has been especially true in our Linux runs.  I was able to reproduce this locally and figured out that this is due to when the `spinValue` is checked on the server.  Most of the time it is checked on the stream packet that creates the new stream on the listener and this is perfect, it should always contain the right value then.  However, some of the time, there is a delay and the `spinValue` does not get checked until the client sends the ACK to the server for the echo stream frame and this has the spun value on it and makes the test fail.  

So the sequence looks like this:
```
log(packet:coalesced:outbound:) [L1:1] rcv 0.015s SH<0>
log(packet:coalesced:outbound:) [L1:1] ← ACK[[0]]
log(packet:coalesced:outbound:) [L1:1] ← S0[0;12] FIN
append(_:packetNumberSpace:now:) [L1:1] appending pn PacketNumber(0) for space applicationData
(This packet contains the right spin value inbound to the server)

log(packet:coalesced:outbound:) [L1:1] snd 0.003s SH<1>
log(packet:coalesced:outbound:) [L1:1] → ACK[[0]]
log(packet:coalesced:outbound:) [L1:1] → S0[0;12] FIN
(Server ACKs the stream frame and sends another stream frame back to the client with a FIN)

log(packet:coalesced:outbound:) [C1:1] rcv 0.006s SH<1>
log(packet:coalesced:outbound:) [C1:1] ← ACK[[0]]
log(packet:coalesced:outbound:) [C1:1] ← S0[0;12] FIN
(Client receives the ACK and stream frame with FIN)

log(packet:coalesced:outbound:) [C1:1] snd 0.009s SH<1>
log(packet:coalesced:outbound:) [C1:1] → ACK[[0-1]]
(Client sends the ACK back with the spun value)

log(packet:coalesced:outbound:) [L1:1] rcv 0.011s SH<1>
log(packet:coalesced:outbound:) [L1:1] ← ACK[[0-1]]
append(_:packetNumberSpace:now:) [L1:1] appending pn PacketNumber(1) for space applicationData
(Server gets the spun value and the unit test checks the server at this point and makes the test fail)
```
The solution is to add another path recognizing that the `spinValue` did make it to the server successfully.  After adding this approach I can run the test 1000 times successfully.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/QUICPath.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICPath.swift
@@ -188,7 +188,6 @@ public final class QUICPath: MultiplexingDatagramPath<QUICConnection>, Equatable
         static let isFlowControlled = Flags(rawValue: 1 << 9)
         static let l4sEnabled = Flags(rawValue: 1 << 10)
         static let reportedIdleEvent = Flags(rawValue: 1 << 11)
-        static let spinValueForTestingSeen = Flags(rawValue: 1 << 12)
     }
     private var flags = Flags()
 
@@ -241,10 +240,6 @@ public final class QUICPath: MultiplexingDatagramPath<QUICConnection>, Equatable
     var reportedIdleEvent: Bool {
         get { flags.contains(.reportedIdleEvent) }
         set { if newValue { flags.insert(.reportedIdleEvent) } else { flags.remove(.reportedIdleEvent) } }
-    }
-    var spinValueForTestingSeen: Bool {
-        get { flags.contains(.spinValueForTestingSeen) }
-        set { if newValue { flags.insert(.spinValueForTestingSeen) } else { flags.remove(.spinValueForTestingSeen) } }
     }
 
     var log: LogPrefixer {
--- a/Sources/SwiftNetwork/QUIC/QUICPath.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICPath.swift
@@ -188,6 +188,7 @@ public final class QUICPath: MultiplexingDatagramPath<QUICConnection>, Equatable
         static let isFlowControlled = Flags(rawValue: 1 << 9)
         static let l4sEnabled = Flags(rawValue: 1 << 10)
         static let reportedIdleEvent = Flags(rawValue: 1 << 11)
+        static let spinValueForTestingSeen = Flags(rawValue: 1 << 12)
     }
     private var flags = Flags()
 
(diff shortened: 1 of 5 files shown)


Sources/SwiftNetwork/QUIC/QUICPath.swift
```diff
@@ -188,6 +188,7 @@ public final class QUICPath: MultiplexingDatagramPath<QUICConnection>, Equatable
         static let isFlowControlled = Flags(rawValue: 1 << 9)
         static let l4sEnabled = Flags(rawValue: 1 << 10)
         static let reportedIdleEvent = Flags(rawValue: 1 << 11)
+        static let spinValueForTestingSeen = Flags(rawValue: 1 << 12)
```
tfpauly: I'm a bit concerned about putting in more things that the test code would directly reach into the objects for (instead of grabbing from the metadata APIs, etc).

Could we instead maybe have an approach where the test looks at the spin bit on the packets going back and forth through the bridge protocol?

agnosticdev: Observing from the bridge protocol is a bit more invasive, but since it's only used for testing it probably not an issue.  Added a mechanism to sift outbound short header packets via a callback and check the first byte for the Spin Bit. [2f32370](https://github.com/apple/swift-network-evolution/pull/41/commits/2f32370f8e43309d4c834eccbf75e8253555716b)

Tests/SwiftNetworkTests/QUICTestHarness.swift
```diff
@@ -76,6 +76,9 @@ final class QUICTestHarness {
 
         let clientInstance: QUICProtocol.Instance
         let serverInstance: QUICProtocol.Instance
+
+        let clientBridge: BridgeDatagramProtocol.Instance
```
tfpauly: The bridge instance should not be held directly, and we shouldn't configure any blocks on it directly. Please pass the block in the ProtocolOptions like other options for the bridge protocol and all others.

tfpauly (changes requested):

Tests/SwiftNetworkTests/QUICTestHarness.swift
agnosticdev: Moved the observation block to `BridgeOptions`.  Addressed in [2c404e9](https://github.com/apple/swift-network-evolution/pull/41/commits/2c404e925bcf969f6a8ef2c2a012bc4fe771427e)

tfpauly (approved):
Thanks for the updates!

rpaulo (approved):

PR #44, opened 2026-07-06
Add CIDR and domain pattern matching as utilities

kpodosin:
Adds to Utilities/IPAddress+CIDR: 
- IPv4 and IPv6 CIDR subnet matching
- right-to-left domain pattern matching with wildcard and suffix support
- NAT64 synthesis detection on IPv6Address

Adds tests for above in SwiftNetworkCIDRTests

### DIFF
--- a/Sources/SwiftNetwork/Utilities/IPAddress+CIDR.swift
+++ b/Sources/SwiftNetwork/Utilities/IPAddress+CIDR.swift
@@ -1,0 +1,178 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+// MARK: - CIDR parsing
+
+// Parses an IPv4 CIDR string into a masked network address and subnet mask in network byte order.
+// Supports shorthand notation (e.g. "17.142/16" expands to "17.142.0.0/16").
+private func parseCIDRv4(_ cidr: String) -> (network: UInt32, mask: UInt32)? {
+    let parts = cidr.split(separator: "/", maxSplits: 1, omittingEmptySubsequences: false)
+    guard parts.count == 2,
+        let prefixLen = Int(parts[1]), prefixLen >= 0, prefixLen <= 32
+    else { return nil }
+
+    // Expand shorthand notation: "17.142" -> "17.142.0.0", "10" -> "10.0.0.0"
+    var addrString = String(parts[0])
+    let dotCount = addrString.count(where: { $0 == "." })
+    if dotCount < 3 {
+        addrString += String(repeating: ".0", count: 3 - dotCount)
+    }
+
+    guard let addr = IPv4Address(addrString) else { return nil }
+    let hostMask: UInt32 = prefixLen == 0 ? 0 : UInt32.max << UInt32(32 - prefixLen)  // shift by 32 is undefined
+    let mask = hostMask.bigEndian
+    return (network: addr.addressValue & mask, mask: mask)
+}
+
+// Parses an IPv6 CIDR string into a masked network address and subnet mask,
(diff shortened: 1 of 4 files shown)


Sources/SwiftNetwork/Utilities/IPAddress+CIDR.swift
```diff
@@ -0,0 +1,197 @@
...
+        #endif
+        return matchesDomainPattern(debugDescription, pattern: pattern)
+    }
+}
+
+@available(Network 0.1.0, *)
+extension Endpoint {
+    /// Returns true if this endpoint matches `pattern`. `"*"` matches all endpoints. Host endpoints
+    /// are matched by hostname; address endpoints are matched by IP address or CIDR block.
+    func matchesProxyPattern(_ pattern: String) -> Bool {
```
tfpauly: Probably this should be just "matchesPattern", not specific to proxies?

Sources/SwiftNetwork/Utilities/IPAddress+CIDR.swift
```diff
@@ -0,0 +1,197 @@
...
+
+    // Expand shorthand notation: "17.142" -> "17.142.0.0", "10" -> "10.0.0.0"
+    var addrString = String(parts[0])
+    let dotCount = addrString.count(where: { $0 == "." })
+    if dotCount < 3 {
+        addrString += String(repeating: ".0", count: 3 - dotCount)
+    }
+
+    var inAddr = in_addr()
+    guard inet_pton(AF_INET, addrString, &inAddr) == 1 else { return nil }
```
tfpauly: Let's remove any use of inet_pton, in_addr, etc. We should make this pure swift and standalone, with no platform dependencies.

Sources/SwiftNetwork/Utilities/IPAddress+CIDR.swift
```diff
@@ -0,0 +1,197 @@
...
+)? {
+    let parts = cidr.split(separator: "/", maxSplits: 1, omittingEmptySubsequences: false)
+    guard parts.count == 2,
+        let prefixLen = Int(parts[1]), prefixLen >= 0, prefixLen <= 128
+    else { return nil }
+
+    var in6Addr = in6_addr()
+    guard inet_pton(AF_INET6, String(parts[0]), &in6Addr) == 1 else { return nil }
+
+    let rawNet = withUnsafeBytes(of: in6Addr) {
```
tfpauly: Let's avoid unsafe

Sources/SwiftNetwork/Utilities/IPAddress+CIDR.swift
```diff
@@ -0,0 +1,197 @@
...
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif !NETWORK_STANDALONE
+import Darwin
+#endif
+
+// MARK: - CIDR parsing
+
+#if !NETWORK_STANDALONE
```
tfpauly: Let's remove this check and not rely on inet_ functions

tfpauly (approved):

Sources/SwiftNetwork/Utilities/IPAddress+CIDR.swift
```diff
@@ -0,0 +1,174 @@
...
+/// (`*.apple.com`). Both inputs are case-insensitive; trailing dots are stripped before matching.
+func matchesDomainPattern(_ string: String, pattern: String) -> Bool {
+    let host = (string.hasSuffix(".") ? String(string.dropLast()) : string).lowercased()
+    let pat = (pattern.hasSuffix(".") ? String(pattern.dropLast()) : pattern).lowercased()
+    if host == pat { return true }
+    let hostNodes = host.split(separator: ".", omittingEmptySubsequences: false)
+    var patNodes = pat.split(separator: ".", omittingEmptySubsequences: false)
+    // A leading empty segment (from a pattern starting with ".", e.g. ".apple.com")
+    // is treated as a wildcard, matching like "*.apple.com".
+    if patNodes.first?.isEmpty == true { patNodes[0] = "*" }
```
agnosticdev, line 88: Is this piece correct? If we pass in a "." does that either fail silently here or default to "*" here?

kpodosin: Yes, that is expected to default to *

agnosticdev (approved):

PR #46, opened 2026-07-09
SwiftQUIC: Expose remote datagram frame size through connection metadata

agnosticdev:
Expose remote datagram frame size through connection metadata for API consumption.

### DIFF
--- a/Sources/SwiftNetwork/Protocols/QUICConnectionProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/QUICConnectionProtocol.swift
@@ -753,6 +753,7 @@ public struct QUICConnectionProtocol: NetworkProtocol {
         public var applicationErrorReason: String?
 
         public var activeConnectionIDLimit: Int = 0
+        public var remoteMaxDatagramFrameSize: UInt16 = 0
         #if NETWORK_PRIVATE
         var privateStorage = QUICConnectionMetadataPrivateStorage()
         #endif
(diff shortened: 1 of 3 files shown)


Sources/SwiftNetwork/Protocols/QUICConnectionProtocol.swift
```diff
@@ -1093,6 +1094,15 @@ public struct QUICConnectionProtocol: NetworkProtocol {
             }
         }
 
+        public func getRemoteDatagramFrameSize() -> UInt16 {
```
tfpauly: The value for this transport parameter can't change so I wonder if it's better/simpler to model on activeConnectionIDLimit, and just set it once — no locks, no callbacks, etc

tfpauly (changes requested):

Sources/SwiftNetwork/Protocols/QUICConnectionProtocol.swift
agnosticdev: I think there is value in this approach too, but this starts to fragment the current pattern we have today.

agnosticdev: Addressed in [72a6981](https://github.com/apple/swift-network-evolution/pull/46/commits/72a6981f274e69863ad98b8cb43f99e52dc1b7bf)

tfpauly (approved):

PR #49, opened 2026-07-14
Don't gate PTO packets on address validation

rnro:
At the moment the code will not send an ack-eliciting ping once the peer's address is validated. This appears to be linked to RFC 9000 Section 8.1, which states
> To prevent this deadlock, clients MUST send a packet on a Probe Timeout (PTO) [...] the client MUST send an Initial packet in a UDP datagram that contains at least 1200 bytes if it does not have Handshake keys, and otherwise send a Handshake packet.

However I think this is a slight mis-read, and that RFC 9002 Section 6.2.4
> When a PTO timer expires, a sender MUST send at least one ack-eliciting packet [...] When there is no data to send, the sender SHOULD send a PING or other ack-eliciting frame in a single packet, rearming the PTO timer.

shows that whilst validation is an important inflection point for the two behaviors, PTO pings should still be sent after it. The pseudocode in A9 I think also indicates this.

A new test with the current code results in outstanding frames which will never have a PTO fire and the current code will end up in a tight loop.

This change removes the validation distinction in `sendPTO`, along with the now-unused validation local and the timer-clear branch it fed; the difference in behavior required by the RFC will be enacted by the timer.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/Recovery.swift
+++ b/Sources/SwiftNetwork/QUIC/Recovery.swift
@@ -656,14 +656,16 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
             _ packets: consuming NetworkUniqueDeque<SentPacketRecord>,
             connection: QUICConnection
         ) -> Bool {
-            var packets = packets
-            guard !packets.isEmpty else {
+            if packets.isEmpty {
                 return false
             }
+
+            var packets = packets
             while !packets.isEmpty {
-                let packet = packets.remove(at: 0)
+                let packet = packets.removeFirst()
                 sentPacket(packet, time: connection.now, connection: connection)
             }
+
             return true
         }
     }
@@ -754,7 +756,7 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
     ) {
         var sentPackets = sentPackets
         while !sentPackets.isEmpty {
-            let packet = sentPackets.remove(at: 0)
+            let packet = sentPackets.removeFirst()
             sentPacket(packet, time: connection.now, connection: connection)
         }
         if inBatch {
@@ -1115,25 +1117,27 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
 
     mutating func sendPTO(connection: QUICConnection, path: QUICPath) {
         var sentPTO = false
-        let (_, pnSpace) = getEarliestTime(
+
+        let (_, packetNumberSpace) = getEarliestTime(
             earliestTimeType: EarliestTimeType.lastSentAckElicitingTime,
             connection: connection
         )
-        var hasAckEliciting = false
-        connection.withPendingItems(for: pnSpace) { pendingItems in
-            hasAckEliciting = pendingItems.hasAckElicitingPendingItems
-        }
-        let peerCompletedValidation = peerCompletedValidation(connection: connection)
-        var shouldClearTimer = false
+
+        let hasAckEliciting = connection.withPendingItems(for: packetNumberSpace) { $0.hasAckElicitingPendingItems }
         var discardInitialRecoveryState = false
         applyToAllInnerStatesMutable { innerState, packetNumberSpace in
             let ackElicitingPacketsInFlight = innerState.ackElicitingPacketsInFlight
-            if ackElicitingPacketsInFlight == 0 {
+            guard ackElicitingPacketsInFlight > 0 else {
+                if _slowPath(ackElicitingPacketsInFlight < 0) {
+                    connection.log.fault("ackElicitingPacketsInFlight negative: \(ackElicitingPacketsInFlight)")
+                }
                 return
             }
+
             connection.log.datapath(
                 "PTO \(path.recoveryState.PTOCount) (\(packetNumberSpace)) fired on path \(path.identifier) with \(ackElicitingPacketsInFlight) ack-eliciting packets in flight"
             )
+
             if hasAckEliciting {
                 connection.log.datapath("Sending next frames with new data as PTOs")
                 sentPTO = true
@@ -1147,8 +1151,10 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
                         "Unable to force send PTOs, likely flow-controlled or unavailable"
                     )
                 }
-            } else if ackElicitingPacketsInFlight > 0 {
+
+            } else {
                 connection.log.datapath("Retransmitting two tail-packets as PTO")
+
                 var addedPackets = 0
                 let packetCount = innerState.outstandingPackets.count
                 for i in 0..<packetCount {
@@ -1174,6 +1180,7 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
                             discardInitialRecoveryState: &discardInitialRecoveryState
                         )
                     }
+
                     if let packets, innerState.recordSentPackets(packets, connection: connection) {
                         sentPTO = true
                         addedPackets += 1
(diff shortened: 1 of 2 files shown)


agnosticdev:
> At the moment the code will not send an ack-eliciting ping once the peer's address is validated. 
> 
> PTO pings should still be sent after it.

PTO can generate PINGs still if the ACK blocks get large enough to piggyback a PING frame to make them ACK-eliciting, see the code here: https://github.com/apple/swift-network-evolution/blob/main/Sources/SwiftNetwork/QUIC/Ack.swift#L304C13-L304C24

Sources/SwiftNetwork/QUIC/Recovery.swift
```diff
@@ -1115,25 +1117,27 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
...
-        var shouldClearTimer = false
+
+        let hasAckEliciting = connection.withPendingItems(for: packetNumberSpace) { $0.hasAckElicitingPendingItems }
         var discardInitialRecoveryState = false
         applyToAllInnerStatesMutable { innerState, packetNumberSpace in
             let ackElicitingPacketsInFlight = innerState.ackElicitingPacketsInFlight
-            if ackElicitingPacketsInFlight == 0 {
+            guard ackElicitingPacketsInFlight > 0 else {
+                if _slowPath(ackElicitingPacketsInFlight < 0) {
+                    connection.log.fault("ackElicitingPacketsInFlight negative: \(ackElicitingPacketsInFlight)")
```
agnosticdev: We should probably make sure we are not underflowing `ackElicitingPacketsInFlight`, possibly adding an underflow check to `-= 1` since this is an integer.  That way we do not have to log out these negative values.

Sources/SwiftNetwork/QUIC/Recovery.swift
```diff
@@ -1188,37 +1195,33 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
             }
         }
 
+        // Anti deadlock PING frame (i.e PADDED PING). The PING will be padded when we send an initial packet.
+        // Issue whether or not handshake has completed, the timer will make the distinction.
         if !sentPTO {
             connection.log.datapath("Sending a PING as PTO")
-            if peerCompletedValidation {
-                connection.log.fault("PTO fired after validation")
```
agnosticdev: Historically this log has been a warning that something in Recovery may not be working as expected, or you may have run into a race condition in the logic.  For example, I am looking at a case now where ackElicitingPacketsInFlight is 1, but then the handshake completes, clears the ack-eliciting packets, resetPTOCount set the count back 0, and then sendPTO falls into this path as a matter of timing.  When this happens `sendPTO` tries to reset the timer and realizes that there are now no ackElicitingPacketsInFlight and then cancels the timer and continues on as normal.
So in this path we have now removed the resetTimer functionality, what is the rationale there?

rpaulo:
As discussed with @agnosticdev, the code was assuming that if we hadn't already finished peer validation (i.e., we were in the middle of the handshake), we should have had ack eliciting packets to send, but it seems like that may not be true when the timer is scheduled in conjunction with the receipt of an ACK.

rnro:
> > At the moment the code will not send an ack-eliciting ping once the peer's address is validated.
> > PTO pings should still be sent after it.
> 
> PTO can generate PINGs still if the ACK blocks get large enough to piggyback a PING frame to make them ACK-eliciting, see the code here: https://github.com/apple/swift-network-evolution/blob/main/Sources/SwiftNetwork/QUIC/Ack.swift#L304C13-L304C24

Yep, my initial wording was too strong here. They can be sent in that case, I meant this last-resort PING. That highlighted case only fires a PING for specific traffic patterns - it needs inbound traffic gappy enough to produce more than 5 ACK blocks, and the flag is consumed on write. With no inbound traffic this reproduces deterministically.

Sources/SwiftNetwork/QUIC/Recovery.swift
rnro: I don't think we have lost the timer. By my read `sendPTO`'s only caller is `timerFired`, which calls `resetTimer` right after it returns, and that cancels when nothing is in flight and the peer has validated us. So I think the existing `shouldClearTimer` cancel was redundant.

I think your callout about the potential race is right and I regressed that here. With nothing in flight the loop does nothing, so the unconditional fallback would have sent a spurious PING. It's now gated on whether anything is in flight rather than on validation. That case keeps the fault and sends nothing, and the probe is only unconditional when there's something to probe for.

rnro:
I've pushed a follow-up commit for the review comments, plus two issues they turned up, and merged in main.

Both `ackElicitingPacketsInFlight` decrement sites now refuse to go below zero, so nothing logs negative
values.

The PTO fired after validation path is no longer unconditional. It's gated on whether anything is
ack-eliciting in flight. The "state changed between arming the timer and it firing" case keeps its old
behaviour — fault, send nothing, resetTimer cancels. The probe stays unconditional when there really are
packets in flight. Those are two separate RFC obligations and I believe both are covered: a probe when packets are
in flight (9002 6.2.4), and the padded anti-deadlock packet before validation (6.2.2.1).

The other issues I noticed were in the same function. 
* `sentPTO` was set before the packet was built, so a pending item that writes no payload 
(e.g. a stream queued for service whose flow has since been torn down) incorrectly counted as a
probe and suppressed the PING.
* `hasAckElicitingPendingItems` was read for the wrong space. It was evaluated *once* for a 
particular packet number space (from `getEarliestTime`) and then applied to all three spaces 
but I think pending items are per-space.

I checked each test fails only for its own defect, by reverting each fix separately.

The merge conflicted with #62 ("Stop logging a fault for valid PTO cases") and I resolved in favour of
this PR. I think this PR was rightly removing fault logging from a case which can legitimately arise, 
but I don't think sending nothing is the right action. By my read RFC 9002 says we should send a PING: 
6.2.4 requires an ack-eliciting probe, 7.5 says probes must not be blocked by congestion control, and 
a PING isn't flow controlled.

Sources/SwiftNetwork/QUIC/Recovery.swift
```diff
@@ -1190,47 +1200,43 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
         }
 
         if !sentPTO {
+            if totalAckElicitingPacketsInFlight == 0, peerCompletedValidation(connection: connection) {
+                // Nothing is in flight to probe for, so the state must have changed between arming the
+                // timer and it firing (e.g. the handshake completed and cleared the in-flight packets).
```
agnosticdev, line 1203: Yep, thank you for calling this condition out.

agnosticdev (approved):

rpaulo:
> > At the moment the code will not send an ack-eliciting ping once the peer's address is validated.
> > PTO pings should still be sent after it.
> 
> PTO can generate PINGs still if the ACK blocks get large enough to piggyback a PING frame to make them ACK-eliciting, see the code here: https://github.com/apple/swift-network-evolution/blob/main/Sources/SwiftNetwork/QUIC/Ack.swift#L304C13-L304C24

I think these two things are orthogonal.  PTOs happen when a timer fire.  Adding a PING frame to an ACK happens when we send an ACK, which is usually when we receive packets or when the delayed ACK timer fires.

rpaulo (approved):

Sources/SwiftNetwork/QUIC/Recovery.swift
```diff
@@ -656,14 +660,16 @@ struct Recovery: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
...
             connection: QUICConnection
         ) -> Bool {
-            var packets = packets
-            guard !packets.isEmpty else {
+            if packets.isEmpty {
                 return false
             }
+
+            var packets = packets
             while !packets.isEmpty {
```
tfpauly: This could probably be simplified to:
```
while let packet = packets.popFirst() {
```

tfpauly (approved):

Sources/SwiftNetwork/QUIC/Recovery.swift
rnro: Ah right, yeah, I was shadowing it so it was mutable but that's not needed with `consuming`.

kkuk24 (approved):

PR #51, opened 2026-07-14
Add Reassembly support for Swift IPv6

agnosticdev:
This change adds reassembly support for inbound IPv6 fragments. 
This change only supports inbound fragmentation and reassembly.  Outbound fragmentation will come in a later change.
This change takes a similar approach to IPv4 reassembly support, if no fragments are detected in `processInboundFrames` it will take the fast path and do nothing with reassembly.

### DIFF
--- a/Sources/SwiftNetwork/Protocols/Frame.swift
+++ b/Sources/SwiftNetwork/Protocols/Frame.swift
@@ -576,7 +576,7 @@ public struct Frame: ~Copyable {
                 }
                 return
             }
-            guard newValue < 2 ^ 6 else {
+            guard newValue < 64 else {
                 Logger.proto.fault("Cannot set DSCP value of \(newValue)")
                 return
             }
--- a/Sources/SwiftNetwork/Protocols/IPProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/IPProtocol.swift
@@ -1018,7 +1018,7 @@ public struct IPProtocol: NetworkProtocol {
             static let hopByHopExtensionHeader: UInt8 = 0
             static let routingExtensionHeader: UInt8 = 43
             static let destinationOptionsExtensionHeader: UInt8 = 60
-            static let fragmentExtensionHeaderLength: Int = 8
+            static let fragmentExtensionHeaderLength = 8
             static let ip6fOffMask: UInt16 = 0xFFF8
             static let ip6fMoreFragmentMask: UInt16 = 0x0001
 
--- a/Sources/SwiftNetwork/Protocols/IPProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/IPProtocol.swift
@@ -371,11 +371,6 @@ public struct IPProtocol: NetworkProtocol {
             var dscpValue: UInt8?
         }
 
-        struct IPReassemblyState: ~Copyable {
-            var reassemblyID: UInt16
-            var inputReassemblyFrames = FrameArray()
-        }
-
         struct IPInstanceFlags: OptionSet {
             init(rawValue: Self.RawValue) {
                 self.rawValue = rawValue
@@ -448,7 +443,12 @@ public struct IPProtocol: NetworkProtocol {
             var flags = IPInstanceFlags()
             var counters = IPCounters()
             var pathProperties = IPPathProperties()
-            var reassemblyState: IPReassemblyState?
+            var reassemblyState: IPv4ReassemblyState?
+
+            struct IPv4ReassemblyState: ~Copyable {
+                var reassemblyID: UInt16
+                var inputReassemblyFrames = FrameArray()
+            }
 
             static var headerLength: Int {
                 MemoryLayout<UInt32>.size * 5
@@ -664,7 +664,7 @@ public struct IPProtocol: NetworkProtocol {
                 // Only update the stored reassembly ID when processing a real fragment and not on force flush
                 if !forceFlush {
                     if reassemblyState == nil {
-                        reassemblyState = IPReassemblyState(reassemblyID: ipID)
+                        reassemblyState = IPv4ReassemblyState(reassemblyID: ipID)
                     } else {
                         reassemblyState?.reassemblyID = ipID
                     }
@@ -672,7 +672,6 @@ public struct IPProtocol: NetworkProtocol {
             }
 
             mutating func processInboundFrames(_ log: borrowing NetworkLoggerState, _ inboundFrames: inout FrameArray) {
-                // Hoist loop-invariant values.
                 let localAddress: UInt32 = self.localAddress.addressValue
                 let remoteAddress: UInt32 = self.remoteAddress.addressValue
                 let mask = (0xF000_0000 as UInt32).bigEndian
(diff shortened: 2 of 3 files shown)


Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1013,7 +1012,28 @@ public struct IPProtocol: NetworkProtocol {
             var flags = IPInstanceFlags()
             var counters = IPCounters()
             var pathProperties = IPPathProperties()
-            var reassemblyState: IPReassemblyState?
+            var reassemblyState: IPv6ReassemblyState?
+
+            static let fragmentExtensionHeader: UInt8 = 44
+            static let hopoptsExtensionHeader: UInt8 = 0
```
rpaulo:
```suggestion
            static let hopByHopExtensionHeader: UInt8 = 0
```

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1013,7 +1012,28 @@ public struct IPProtocol: NetworkProtocol {
...
+            static let routingExtensionHeader: UInt8 = 43
+            static let dstoptsExtensionHeader: UInt8 = 60
```
rpaulo:
```suggestion
            static let destinationOptionsExtensionHeader: UInt8 = 60
```

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1030,17 +1050,269 @@ public struct IPProtocol: NetworkProtocol {
...
+                var payloadLength: UInt16 = 0
+                var firstProto: UInt8 = 0
+                var fragmentOffset: UInt16 = 0
+                var moreFragments = false
+                var headerOffset = IPv6Instance.headerLength
+                var nextProtocol: UInt8 = 0
+                var foundFragment = false
+
+                let result = Deserializer.deserialize(&frame, claim: false) { read throws(DeserializationError) in
+                    try read.skip(4)
```
rpaulo, line 1066: Shouldn't the first 4 bytes still be validated?

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1030,17 +1050,269 @@ public struct IPProtocol: NetworkProtocol {
...
+                    try read.uint16NetworkByteOrder(&payloadLength)
+                    try read.uint8(&firstProto)
+                    try read.skip(1)
+                    try read.skip(32)  // src and dst addresses
```
rpaulo, line 1070: Same question: is this being validated anywhere else?

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1030,17 +1050,269 @@ public struct IPProtocol: NetworkProtocol {
...
+                        }
+                        if dropped > 0 {
+                            log.error(
+                                "Dropping \(dropped) incomplete IPv6 fragments for ID \(reassemblyState?.reassemblyID ?? 0)"
+                            )
+                        }
+                    }
+                }
+                if !forceFlush {
+                    if reassemblyState == nil {
```
rpaulo, line 1292: You could revert this and do `if let reassemblyState = reassemblyState {` and then you avoid `reassemblyState?`.

rpaulo (approved):
Approved pending minor comments.

Sources/SwiftNetwork/Protocols/IPProtocol.swift
agnosticdev: Addressed in [2629238](https://github.com/apple/swift-network-evolution/pull/51/commits/2629238ce9fef827f52356eaa2c166ece9f194d6)

Sources/SwiftNetwork/Protocols/IPProtocol.swift
agnosticdev: Addressed in [2629238](https://github.com/apple/swift-network-evolution/pull/51/commits/2629238ce9fef827f52356eaa2c166ece9f194d6)

Sources/SwiftNetwork/Protocols/IPProtocol.swift
agnosticdev: This would be redundant because the version / flow label was already validated in `processInboundFrames`, that is why its skipped here when the frame is parsed from the reassembledState.

Sources/SwiftNetwork/Protocols/IPProtocol.swift
agnosticdev: Same here, the src and dest addresses were validated in `processInboundFrames`.

Sources/SwiftNetwork/Protocols/IPProtocol.swift
agnosticdev: This would not work because `reassemblyState` is a non-copyable struct and setting it locally would cause ownership issues.

rpaulo: You can do it with this sort of trick:

```
switch reassemblyState {
case .some(let state): 
// do stuff with state which is now non nil reassemblyState
case .none:
// nil reassemblyState
}
```

Or you can even do `if .case(let state) = reassemblyState` tricks.

agnosticdev: This trick does work for setting a property such as `.reassemblyID`, but not when creating a new `IPv6ReassemblyState` object.  You would run into a partially consuming self issue in that case I believe and would have to devise a strategy for assigning the entire instance to itself again, i.e., `.ipv6(instance)`.
This would also incur extra cycles we don't need to expend, leaving this as-is.

Sources/SwiftNetwork/Protocols/Frame.swift
```diff
@@ -576,7 +576,7 @@ public struct Frame: ~Copyable {
                 }
                 return
             }
-            guard newValue < 2 ^ 6 else {
+            guard newValue < 64 else {
```
tfpauly, line 579: Why this change? It's the same value

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1013,7 +1012,28 @@ public struct IPProtocol: NetworkProtocol {
...
             var counters = IPCounters()
             var pathProperties = IPPathProperties()
-            var reassemblyState: IPReassemblyState?
+            var reassemblyState: IPv6ReassemblyState?
+
+            static let fragmentExtensionHeader: UInt8 = 44
+            static let hopByHopExtensionHeader: UInt8 = 0
+            static let routingExtensionHeader: UInt8 = 43
+            static let destinationOptionsExtensionHeader: UInt8 = 60
+            static let fragmentExtensionHeaderLength: Int = 8
```
tfpauly:
```suggestion
            static let fragmentExtensionHeaderLength = 8
```

agnosticdev: Addressed in [e1dd861](https://github.com/apple/swift-network-evolution/pull/51/commits/e1dd861c068bcada2b134b2333fd4659440ec609)

Sources/SwiftNetwork/Protocols/Frame.swift
agnosticdev: This is an interesting case; when running the test `testIPEcho` we would receive errors for DSCP values less than 50 in IPv6 cases.  This line seemed to be doing bitwise XOR instead of calculating 2 to the power of 6, which is what I think was meant here.  [Reference on bitwise XOR](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/advancedoperators/#Bitwise-XOR-Operator).

tfpauly: Ohhh good call then. Thanks for fixing.

tfpauly (approved):

PR #52, opened 2026-07-15
Increase timeout for testQUICEcho40KiBMultistreamWithMetrics

agnosticdev:
Increase the metric fetching timeout to provide more leeway when multiple tests are running at one time.

### DIFF
--- a/Tests/SwiftNetworkTests/QUICTestHarness.swift
+++ b/Tests/SwiftNetworkTests/QUICTestHarness.swift
@@ -1118,7 +1118,7 @@ final class QUICTestHarness {
                     XCTAssertNotNil(serverReports)
                     snapshotExpectation.fulfill()
                 }
-                wait(for: [snapshotExpectation], timeout: 2.0)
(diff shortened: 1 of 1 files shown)


ekinnear:
Do we have evidence that this would have worked in 5 seconds vs. fixing it by making it faster?

agnosticdev:
> Do we have evidence that this would have worked in 5 seconds vs. fixing it by making it faster?

This has been seen only in internal test systems and I am not able to reproduce this.  What I am wanting to do is increase the timeout here to detect if there is actually a logical problem taking place or if we are running into a timeout that was too low when the system is running concurrent tests.  For example, for 2 seconds, if the system is having an issue or running tests concurrently then I can see where this could randomly timeout un load.  I have a hard time believing that the same would be true for 5 seconds.  If we still see the same random failures at 5 seconds then it's time to rewrite this logic in our test harness.

rpaulo (approved):

tfpauly (approved):

PR #55, opened 2026-07-18
SwiftQUIC: Provide path events when QUICPath changes or is validated

agnosticdev:
There is a need for swift-nio-quic to observe events when `QUICPath` changes or becomes validated. This typically happens during connection migration.  It is not appropriate to expose `QUICPath` directly so this change exposes a new type called `QUICPathInfo` that carries local and remote Endpoint information about the new datagram path underneath QUIC when it changes.  When `QUICPath` changes or becomes validated `QUIC` checks with the lower protocol (UDP) to obtain the endpoints of the new path below it and sends a `QUICEvent` that can be observed at the connection level through an object such as `NewStreamFlowHarness` and the application level can update needed info based on these events.

Design considerations: `QUICPathInfo` needs to be `Sendable` because `DomainSpecificNetworkProtocolEvent` is `Sendable`.  `Endpoint` is not `Sendable` and so I had to expose it through a computed property here.   That is the need for `PathAddress` on `QUICPathInfo`.

### DIFF
--- a/Sources/SwiftNetwork/Protocols/NetworkEvents.swift
+++ b/Sources/SwiftNetwork/Protocols/NetworkEvents.swift
@@ -22,15 +22,9 @@ public struct NetworkEventDomain: Sendable, Hashable, CustomStringConvertible {
 @_spi(ProtocolProvider)
 @available(Network 0.1.0, *)
 public struct QUICPathInfo: Sendable, Equatable {
-    private var datagramPathEndpoints: DatagramPathEndpoints
     public let isValidated: Bool
-    public var remote: Endpoint { datagramPathEndpoints.remote.endpoint }
-    public var local: Endpoint { datagramPathEndpoints.local.endpoint }
-
-    init(datagramPathEndpoints: DatagramPathEndpoints, isValidated: Bool) {
-        self.datagramPathEndpoints = datagramPathEndpoints
-        self.isValidated = isValidated
-    }
+    public let remote: AddressEndpoint
+    public let local: AddressEndpoint
 }
 
 @_spi(ProtocolProvider)
--- a/Sources/SwiftNetwork/Protocols/NetworkEvents.swift
+++ b/Sources/SwiftNetwork/Protocols/NetworkEvents.swift
@@ -22,27 +22,13 @@ public struct NetworkEventDomain: Sendable, Hashable, CustomStringConvertible {
 @_spi(ProtocolProvider)
 @available(Network 0.1.0, *)
 public struct QUICPathInfo: Sendable, Equatable {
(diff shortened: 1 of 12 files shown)


agnosticdev:
Another design consideration for swift-nio-quic is that this change takes into account a defined UDP and IP stack running underneath QUIC.  In swift-nio-quic, QUIC sits on top of a UDP socket and so we will need to further enhance this change to know when the clients endpoint has changed there also so we can properly expose an event in this case as well.

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
```diff
@@ -19,6 +19,34 @@ public struct NetworkEventDomain: Sendable, Hashable, CustomStringConvertible {
     public var description: String { domain }
 }
 
+@_spi(ProtocolProvider)
+@available(Network 0.1.0, *)
+public struct QUICPathInfo: Sendable, Equatable {
+    public enum PathAddress: Sendable, Equatable {
+        case v4(IPv4Address, port: UInt16)
```
tfpauly: Why are these stored as addresses? I think storing as endpoints would be better than forcing it to be its own tuple of address and port.

Sources/SwiftNetwork/Protocols/ProtocolControlHandlers.swift
```diff
@@ -895,6 +895,18 @@ extension ProtocolInstanceReference {
         }
     }
 
+    public func getPathEndpoints(
```
tfpauly, line 898: How is this supposed to work when we have multiple underlying paths? Which set is returned?

Also, it would be nice to return a struct type instead of a tuple

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
agnosticdev: Ideally these would be stored as an `Endpoint`, or as an `AddressEndpoint`.  However those types are not Sendable and `QUICPathInfo` is required to be `Sendable` because `DomainSpecificNetworkProtocolEvent` is `Sendable`.  So I moved it down to the lowest common `Sendable` type which is an address.

Sources/SwiftNetwork/Protocols/ProtocolControlHandlers.swift
agnosticdev: I did switch a struct in the latest changes [c42008f](https://github.com/apple/swift-network-evolution/pull/55/commits/c42008fe6b4a1d4f58da198f84a85b47736dd2c6).  This change does not take into account multiple underlying paths, only the the particular path that going through the events for `handlePathChanged` and `handlePathChallengeResponse` callbacks.

PR #56, opened 2026-07-22
Swift IPv4 add support for outbound fragments

agnosticdev:
Add support for Swift IPv4 for outbound fragmentation.

### DIFF
--- a/Sources/SwiftNetwork/Protocols/IPProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/IPProtocol.swift
@@ -919,13 +919,13 @@ public struct IPProtocol: NetworkProtocol {
             }
 
             mutating func writeOutboundFrames(_ frames: inout FrameArray) {
-                frames.iterateMutableFrames { frame in
+                var output = FrameArray()
+                while var frame = frames.popFirst() {
                     guard frame.unclaim(fromStart: IPv4Instance.headerLength) else {
                         frame.finalize(success: false)
-                        return .removeFrameAndContinue
+                        continue
                     }
 
-                    let totalLength = UInt16(frame.unclaimedLength)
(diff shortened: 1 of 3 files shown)


Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -919,13 +919,13 @@ public struct IPProtocol: NetworkProtocol {
             }
 
             mutating func writeOutboundFrames(_ frames: inout FrameArray) {
-                frames.iterateMutableFrames { frame in
+                var output = FrameArray()
```
tfpauly: Can we avoid unconditionally creating a new array here? If we don't need fragmentation, I'd rather modify in place

agnosticdev: One way we could do this is create a fragments array at the start of the function.  Then keep `frames.iterateMutableFrames` for the standard path.  If a frame in that array is detected to have a payload over the MTU and we need to split it those fragments are added to the fragment array.  The thing to consider is that those newly generated fragments are then added to the end of the outbound frames array.  This is a slight deviation in ordering that we may need to consider here.

agnosticdev: Added the logic in [b6e417a](https://github.com/apple/swift-network-evolution/pull/56/commits/b6e417a8ec297a95d9eb28615f789b49183279d0)

agnosticdev: Added a change to preserve the ordering [37e1d07](https://github.com/apple/swift-network-evolution/pull/56/commits/37e1d077a5ab5010c8eeb94aae8236929ef96f90)

tfpauly (approved):

PR #57, opened 2026-07-26
SwiftQUIC: Bugfix in inboundShortheader

agnosticdev:
This PR fixes incorrect logic when checking the packet DCID against what should be the path SCID in `handleInboundShortHeader`:
```
packet.destinationConnectionID == path.scid
```
This PR also optimizes the equality check for `QUICConnectionID` to now be constant time.  This is based on a helpful post from Karoy [here](https://forums.swift.org/t/how-to-check-two-array-instances-for-identity-equality-in-constant-time/78792/57).  In the `QUICStreamLoad` benchmark the CPU usage for this equality check no longer even shows up and is completely optimized away.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/QUICConnectionID.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICConnectionID.swift
@@ -169,7 +169,6 @@ public struct QUICConnectionID: Sendable, Equatable, CustomStringConvertible {
             return false
         }
         #if !NETWORK_EMBEDDED
-        // Contstant time equality
         return lhs.connectionIDStorage.span.withUnsafeBytes { left in
             rhs.connectionIDStorage.span.withUnsafeBytes { right in
                 memcmp(left.baseAddress!, right.baseAddress!, 20)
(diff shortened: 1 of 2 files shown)


tfpauly (approved):

Sources/SwiftNetwork/QUIC/QUICConnectionID.swift
```diff
@@ -168,6 +168,15 @@ public struct QUICConnectionID: Sendable, Equatable, CustomStringConvertible {
         guard lhs.actualLength == rhs.actualLength else {
             return false
         }
+        #if !NETWORK_EMBEDDED
+        // Contstant time equality
```
rpaulo: I don't think memcmp does constant time equality, so this comment is misleading.
Constant time equality means that no matter the size of the input, the comparison will always take N cycles, but it doesn't mean that the cycles will be small for a very large input.  memcmp will stop when the first byte changes.

rpaulo (approved):

kukzile (approved):

Sources/SwiftNetwork/QUIC/QUICConnectionID.swift
agnosticdev: [7abbf67](https://github.com/apple/swift-network-evolution/pull/57/commits/7abbf6724c9013bc5170f96eb901ff748b6af182)

PR #58, opened 2026-07-28
Reuse the sent packet deque across sendFrames calls

glbrntt:
Motivation:

Both sendFrames() overloads create a new deque of SentPacketRecord on every call, only to hand its contents to recovery and throw the storage away.

Modifications:

- Store a deque on the QUICConnection and reuse it.
- Take 'sentPackets' inout rather than consuming in Recovery.recordSentPackets. It already drained the storage in place so now the connection can reuse the empty deque.

Result:

- Fewer allocations

### DIFF
--- a/Sources/SwiftNetwork/QUIC/QUICConnection.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICConnection.swift
@@ -2967,6 +2967,9 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
     // Indicates whether the asynchronous send continuation is running or not
     private var asyncSendRunning = false
 
+    // Storage for the packets recorded by a single sendFrames() call.
+    private var sentPackets = NetworkUniqueDeque<SentPacketRecord>(minimumCapacity: 10)
+
     private func capacityForPacketNumberSpace() -> Int {
         let packetNumberSpace = PacketNumberSpace.fromKeyState(keyState: self.keyState)
         if packetNumberSpace == .applicationData {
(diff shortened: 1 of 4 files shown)


glbrntt:
When running some http/3 benchmarks (100k requests, each a 1k download with 100 concurrent streams on one connection) this saves 100k allocs (only ~2% or so), but reduces total bytes allocated by ~15%.

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -2980,6 +2983,16 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
         }
     }
 
+    private func shrinkSentPacketsIfNecessary() {
+        assert(self.sentPackets.isEmpty)
+        // 'sentPackets' is held onto for the lifetime of the connection. If a send burst grows it
+        // beyond a certain limit then drop the capacity. This avoids bursty traffic bloating memory
+        // indefinitely.
+        if self.sentPackets.capacity > 512 {
```
rpaulo: This should be a setting in Preferences.swift

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -2998,22 +3011,24 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
             return false
         }
         return withCurrentPath { path in
-            var sentPackets = NetworkUniqueDeque<SentPacketRecord>(minimumCapacity: capacityForPacketNumberSpace())
+            self.sentPackets.reserveCapacity(capacityForPacketNumberSpace())
```
agnosticdev, line 3014: We should profile this and see if this does an explicit reallocation or it ONLY does a reallocation if it needs more capacity. I really don't know what the behavior is here.  If it does a reallocation each time we should get rid of this if we are storing it at the connection level.  Declaring a capacity up front should get us what we want.

glbrntt: It only does a reallocation if there's insufficient capacity, and that path will be very (i.e. compare requested capacity with actual).

This avoids intermediate reallocations if the deque was shrunk in the previous round because it got too big.

agnosticdev: Okay, if it only reallocates on insufficient capacity then we should be good.

agnosticdev (approved):

rpaulo (approved):

PR #64, opened 2026-08-03
Reuse the received frame deque

glbrntt:
Motivation:

Each Packet holds its own deque of parsed frames. The parser also reserved enough capacity for one frame, so every packet allocs at least once and then reallocs if packets contain subsequent frames.

The frames only live between parsing and calling handleInboundPacket, so the storage doesn't need to be tied to a Packet.

Modifications:

- Move the deque of recevied frames from Packet to PacketParser and reserve its capacity once per connection.
- Collapse the two copies of the frame-finalizing drain loop into PacketParser.cleanupReceivedFrames().
- Shrink the deque back down once drained if a packet grew it beyond 128 frames (by default), so that a peer sending a packet of single byte frames can't pin that storage for the lifetime of the connection.

Result:

Fewer allocations

### DIFF
--- a/Sources/SwiftNetwork/QUIC/PacketParser.swift
+++ b/Sources/SwiftNetwork/QUIC/PacketParser.swift
@@ -38,7 +38,7 @@ struct PacketParser: ~Copyable, PrefixedLoggable {
     var framesReceived = NetworkUniqueDeque<QUICFrame>()
 
     // Initial capacity for `framesReceived`.
-    private static var framesReceivedCapacity: Int { 16 }
+    private static var framesReceivedCapacity: Int { 4 }
 
     init(logPrefixer: LogPrefixer) {
         self.log = logPrefixer
--- a/Sources/SwiftNetwork/QUIC/PacketParser.swift
+++ b/Sources/SwiftNetwork/QUIC/PacketParser.swift
@@ -33,8 +33,38 @@ internal import os
 struct PacketParser: ~Copyable, PrefixedLoggable {
     var log: LogPrefixer
 
+    // Temporary storage for the frames parsed out of the packet currently being
+    // processed.
+    var framesReceived = NetworkUniqueDeque<QUICFrame>()
+
+    // Initial capacity for `framesReceived`.
+    private static var framesReceivedCapacity: Int { 16 }
+
     init(logPrefixer: LogPrefixer) {
         self.log = logPrefixer
+        self.framesReceived.reserveCapacity(Self.framesReceivedCapacity)
+    }
+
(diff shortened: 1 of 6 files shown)


glbrntt:
This reduced the runtime of `QUICTransfer` by ~2% and reduced the number of allocations by nearly 5%.

Sources/SwiftNetwork/QUIC/PacketParser.swift
```diff
@@ -342,7 +357,6 @@ struct PacketParser: ~Copyable, PrefixedLoggable {
                 originalLength: originalLength
             )
         }
-        packet.framesReceived.reserveCapacity(1)
```
agnosticdev, line 345: Thank you!  This will show a pretty nice perf impact here.  Last time I measure should be around 133 megacycles.

Sources/SwiftNetwork/QUIC/PacketParser.swift
```diff
@@ -33,8 +33,38 @@ internal import os
 struct PacketParser: ~Copyable, PrefixedLoggable {
     var log: LogPrefixer
 
+    // Temporary storage for the frames parsed out of the packet currently being
+    // processed.
+    var framesReceived = NetworkUniqueDeque<QUICFrame>()
+
+    // Initial capacity for `framesReceived`.
+    private static var framesReceivedCapacity: Int { 16 }
```
rpaulo: This is probably too high because most QUIC packets have only a handful of frames.  What about starting with `4` ?

glbrntt:
```suggestion
    private static var framesReceivedCapacity: Int { 4 }
```

rpaulo (approved):

kkuk24 (approved):

Tests/QUICTests/QUICLayoutTests.swift
```diff
@@ -26,7 +26,7 @@ import XCTest
 final class QUICLayoutTests: XCTestCase {
 
     func testLayoutPacket() {
-        let packetSize = 188
+        let packetSize = 156
```
agnosticdev, line 29: NICE!  Thank you!

agnosticdev (approved):

Tests/QUICTests/QUICLayoutTests.swift
glbrntt: FWIW, and I didn't look at this in any depth, but it looked like `shorthandFrames` may be a candidate for a similar optimization here.

PR #68, opened 2026-08-05
Swift IPv6 Outbound Fragmentation support

agnosticdev:
Adds outbound fragmentation support for Swift IPv6.
IPv6 takes the same approach that IPv4 takes in that it will only attempt to handle fragmentation if needed, otherwise it will use the standard path.

### DIFF
--- a/Sources/SwiftNetwork/Protocols/IPProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/IPProtocol.swift
@@ -920,7 +920,8 @@ public struct IPProtocol: NetworkProtocol {
 
             mutating func writeOutboundFrames(
                 _ frames: inout FrameArray,
-                getDatagramsToSend: (Int, Int) throws -> FrameArray?
+                lower: OutboundDatagramLinkage,
+                selfReference: ProtocolInstanceReference
             ) {
                 frames.iterateMutableFrames { frame in
                     guard frame.unclaim(fromStart: IPv4Instance.headerLength) else {
@@ -978,7 +979,12 @@ public struct IPProtocol: NetworkProtocol {
                             let chunkLength = isLast ? remaining : fragmentRoom
                             // Create the fragment frame with this chunk length
                             let fragmentFrameSize = IPv4Instance.headerLength + chunkLength
-                            guard var allocatedFrames = try? getDatagramsToSend(1, fragmentFrameSize),
+                            guard
+                                var allocatedFrames = try? lower.invokeGetDatagramsToSend(
+                                    selfReference,
+                                    maximumDatagramCount: 1,
+                                    minimumDatagramSize: fragmentFrameSize
+                                ),
                                 var fragmentFrame = allocatedFrames.popFirst()
                             else {
                                 fragmentationSucceeded = false
@@ -1674,7 +1680,8 @@ public struct IPProtocol: NetworkProtocol {
 
             mutating func writeOutboundFrames(
                 _ frames: inout FrameArray,
-                getDatagramsToSend: (Int, Int) throws -> FrameArray?
+                lower: OutboundDatagramLinkage,
+                selfReference: ProtocolInstanceReference
             ) {
                 frames.iterateMutableFrames { (frame: inout Frame) -> FrameArray.FrameIterationResult in
                     _ = frame.unclaim(fromStart: IPv6Instance.headerLength)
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1686,9 +1686,99 @@ public struct IPProtocol: NetworkProtocol {
...
+                        while cursor < payloadLength {
+                            // Determine if last or how large the chunk length is
+                            let remaining = payloadLength - cursor
+                            let isLast = remaining <= fragmentRoom
+                            let chunkLength = isLast ? remaining : fragmentRoom
+                            // Fragment Extension Header + this chunk's payload.
+                            let fragmentLength = UInt16(chunkLength + IPv6Instance.fragmentExtensionHeaderLength)
+                            // Fragment offset flags
+                            let offsetFlags = UInt16(cursor) | (isLast ? 0 : UInt16(IPv6Instance.ip6fMoreFragmentMask))
+                            var fragmentFrame = Frame(count: ipv6CompleteHeaderLength + chunkLength)
```
tfpauly: Hm, this seems concerning — I don't think we can just do this, since we would need to get the frame from the protocol below in cases where the frames aren't allocated to just be unique arrays, but have other backing.

agnosticdev: Yep, that makes sense.  It was a bit more complex than I expected to adopt `getDatagramsToSend` because IPv4 and IPv6 instances cannot call this function directly on the instance.  I added a callback though that gave `writeOutboundFrames` access to this function.  Adopted in [212cfbb](https://github.com/apple/swift-network-evolution/pull/68/commits/212cfbbee1dd0e9f86a2e4a79ac17313af212506)

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -1883,7 +1991,19 @@ public struct IPProtocol: NetworkProtocol {
         }
 
         mutating func sendDatagrams(_ datagrams: consuming FrameArray) throws(NetworkError) {
-            IPInstance.processOutbound(&self.instanceType, datagrams: &datagrams)
+            let lower = self.lower
+            let selfReference = self.effectiveSelfReference
+            IPInstance.processOutbound(
+                &self.instanceType,
+                getDatagramsToSend: { maxCount, minSize in
```
tfpauly: If this is being marked on every IP packet send, I worry that passing a new block every time is too heavy. Why not just pass the reference to `lower` and self reference?

agnosticdev: We should not incur any performance hit here unless the block is actually called.  Switching to using `lower` though is a straightforward change so addressed in [e3cf0f6](https://github.com/apple/swift-network-evolution/pull/68/commits/e3cf0f66f677f9b017f73f6b6b93727b4a6b4dc5)

Sources/SwiftNetwork/Protocols/IPProtocol.swift
```diff
@@ -969,12 +973,23 @@ public struct IPProtocol: NetworkProtocol {
...
-                            // Determine if last or hold large the chunk length is
+                            // Determine if last or how large the chunk length is
                             let remaining = payloadLength - cursor
                             let isLast = remaining <= fragmentRoom
                             let chunkLength = isLast ? remaining : fragmentRoom
                             // Create the fragment frame with this chunk length
-                            var fragmentFrame = Frame(count: IPv4Instance.headerLength + chunkLength)
+                            let fragmentFrameSize = IPv4Instance.headerLength + chunkLength
+                            guard
+                                var allocatedFrames = try? lower.invokeGetDatagramsToSend(
```
tfpauly: Looks good overall. This could probably be optimized by having it batch the get datagrams call to get all of the datagrams it will need if it is more than one, but this is unlikely and probably not a key thing to optimize.

tfpauly (approved):

Sources/SwiftNetwork/Protocols/IPProtocol.swift
agnosticdev: That's a good idea.  Addressed in [b8b408b](https://github.com/apple/swift-network-evolution/pull/68/commits/b8b408b6b65229446de036a4eb633596b38b0c35)

PR #69, opened 2026-08-05
Reduce size of Frame

glbrntt:
Frame is large: 161 bytes with a stride of 168 bytes. A reasonably large part of this is just padding which is easily removed by reordering stored properties.

This patch moves properties around to reduce the size and stride down to 136 bytes.

### DIFF
--- a/Sources/SwiftNetwork/Protocols/Frame.swift
+++ b/Sources/SwiftNetwork/Protocols/Frame.swift
@@ -37,26 +37,34 @@ public struct Frame: ~Copyable {
         case customFinalizer(buffer: UnsafeMutableRawBufferPointer, finalizer: (UnsafeMutableRawBufferPointer) -> Void)
     }
 
-    public var buffer: Buffer
-
     var _bytes: NetworkUniqueArray<UInt8> = .init()
+    var protocolMetadatas: NetworkUniqueDeque<FrameProtocolMetadata> = .init(minimumCapacity: 0)
+    var ipPacketValues: IPPacketValues? = nil
+    public var buffer: Buffer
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/Protocols/Frame.swift
```diff
@@ -37,26 +37,34 @@ public struct Frame: ~Copyable {
...
 
     private var _startOffset: UInt32 = 0
+
+    var timestamp: FrameTimestamp? = nil
+    var flags: Flags = Flags()
+    var _metadataComplete: Bool = false
+    public var connectionComplete: Bool = false
+    private var _endOffset: UInt32 = 0
+    private var _effectiveBufferLength: UInt32 = 0
+    private var _aggregateBufferLength: UInt32 = 0
```
agnosticdev, line 54: So because of the out of order arrangement of these properties extra padding was being added, is that what was happening?

agnosticdev (approved):
Thank you George, just one question but otherwise fine!

Sources/SwiftNetwork/Protocols/Frame.swift
glbrntt: It was the order the properties were in: Swift lays out types in the order they're declared in.

As an example the `Buffer` enum was stored first and is 33 bytes in size with an 8 byte alignment. It was followed by a `NetworkUniqueArray<UInt8>` which is 24 bytes (also 8 byte alignment). Because of the alignment there's 7 bytes of padding between the two, instead you can move smaller types (e.g. 7 booleans) into that space that would otherwise be padding.

tfpauly (approved):

PR #74, opened 2026-08-10
NetworkClock: add manually advanced clock for deterministic tests

rpaulo:
- Replace configurable-initial-value clock with a manual time backing
  store that tests can freeze and advance by hand
- Route `now` and `nowAbsolute` through manual time when set, falling
  back to the system clock otherwise
- Track continuous and absolute time separately, preserving nanosecond
  resolution beyond the system clock's microsecond truncation
- Add tests covering freezing, advancing, and restoring the clock

### DIFF
--- a/Sources/SwiftNetwork/Utilities/NetworkClock.swift
+++ b/Sources/SwiftNetwork/Utilities/NetworkClock.swift
@@ -210,11 +210,12 @@ public struct NetworkDuration: DurationProtocol, Hashable, Equatable, CustomStri
     }
 }
 
-/// A continuous clock with a compact representation and configurable initial value.
+/// A continuous clock with a compact representation that tests can advance manually.
 ///
 /// Mimics `Swift.ContinuousClock`, with two differences:
 /// 1. It uses `NetworkDuration` internally so its size is 8 bytes.
-/// 2. You can create a clock with any value, which is useful for unit tests.
+/// 2. Tests can replace the OS clock with one they advance by hand,
+///    which makes time-dependent behaviour deterministic.
 #if !NETWORK_EMBEDDED
 @_spi(Essentials)
 // Availability due to `SwiftNetwork`'s `System.Time` (used by `Instant.now`)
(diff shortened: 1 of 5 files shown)


Sources/SwiftNetwork/Utilities/NetworkClock.swift
```diff
@@ -224,6 +225,55 @@ public struct NetworkClock: Clock {
     public struct Instant: InstantProtocol, CustomStringConvertible {
         var time: NetworkDuration
 
+        #if !DisableDebugLogging
```
tfpauly: Seems a bit odd to tie this to debug logging — I assume this is just intended to compile stuff out on prod / release builds?

tfpauly (commented):
Very cool! Mainly just questioning if we can use a better availability check.

Sources/SwiftNetwork/Utilities/NetworkClock.swift
rpaulo: Yeah pretty much to avoid the memory cost of adding a class.  The CPU cost of the extra check in now() isn't that bad.

agnosticdev (approved):

Sources/SwiftNetwork/Utilities/NetworkClock.swift
tfpauly: Can we tie it being a debug build or some other trait? I don't think it necessarily makes sense to have this be the logging trait.

rpaulo: We can't tie it to a debug build because we run tests in release mode AFAIK.
I didn't want a trait just for this, that's why I used the debug logging trait which is always enabled during the tests

rnro: I don't think this should be tied to a Trait at all, it should just use a new Define. Traits are used to allow consumers of a package to modify functionality (by enabling a Define in that package). This only impacts our tests, which consumers of a package don't need to care about, so it should just be a local Define.

rpaulo: !DisableDebugLogging is ON by default, that's why I used it.  We can make it a define, but it should be on by default so that we don't have to manually turn something on just to be able to run the tests locally.

rnro: Yeah I think that's fine personally. Toggling it from the default position becomes an optimization.

Tests/SwiftNetworkTests/SwiftNetworkClockTests.swift
```diff
@@ -159,3 +159,123 @@ final class SwiftNetworkClockTests: NetTestCase {
...
+#endif
+
+/// Tests for the manually advanced clock behind `NetworkClock.Instant.now`.
+@available(Network 0.1.0, *)
+final class SwiftNetworkManualClockTests: NetTestCase {
+    private let base = NetworkClock.Instant(milliseconds: 1000)
+
+    override func setUpWithError() throws {
+        #if !NETWORK_INTERNAL_TESTS
+        throw ManualClockUnavailable()
```
tfpauly, line 179: Will this cause just running `swift test` to always fail?

rpaulo: Yes, if you do not compile swift test with NETWORK_INTERNAL_TESTS, it will fail.

tfpauly (approved):

rnro (approved):

PR #75, opened 2026-08-11
Create a NetworkQueue

PayasR:
I noticed we were starting to use dispatch queues and dispatch sources in places throughout the code, creating coupling between this project and libdispatch. This is to abstract out the queue use into a NetworkQueue, which can be backed by a dispatch queue or anything else as needed.


PayasR:
Declining this since the recommendation is to move towards structured concurrency in swift.

PR #76, opened 2026-08-11
Add datagram demux protocol

tfpauly:
Add a small protocol that supports demuxing inbound datagrams based on patterns (such as QUIC connection IDs, etc)

### DIFF
--- a/Sources/SwiftNetwork/Protocols/DemuxProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/DemuxProtocol.swift
@@ -1,0 +1,597 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/Protocols/DemuxProtocol.swift
```diff
@@ -0,0 +1,597 @@
...
+            do { try validate(upper: from, #function) } catch { throw NetworkError.posix(EINVAL) }
+            var requestingIndex: Int? = nil
+
+            let returnArray = serviceInboundFrames(
+                from,
+                maximumDatagramCount: maximumDatagramCount,
+                requestingIndex: &requestingIndex
+            )
+            if let returnArray {
+                return returnArray
```
agnosticdev, line 362: Out of curiosity why does this return here even though there may be entries to match against?  Is it because the `from` does not match?

tfpauly: This is getting frames that are already sorted into the correct upper handler. We only pull new frames from lower when we run out of frames already enqueued.

agnosticdev (approved):

PR #80, opened 2026-08-13
Make endpoint types sendable

porchdog:
Adds Sendability conformance to `Endpoint.EndpointType` enum. This was mostly mechanical save for `BonjourEndpoint` and `ApplicationServiceEndpoint` which needed locking support to protect access to the mutable `Storage` object.

### DIFF
--- a/Sources/SwiftNetwork/Endpoint/ApplicationServiceEndpoint.swift
+++ b/Sources/SwiftNetwork/Endpoint/ApplicationServiceEndpoint.swift
@@ -12,10 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if canImport(Synchronization)
-internal import Synchronization
-#endif
-
 @_spi(Essentials)
 @available(Network 0.1.0, *)
 public struct ApplicationServiceEndpoint: EndpointProtocol, EndpointCommonProtocol, Sendable {
@@ -48,7 +44,7 @@ public struct ApplicationServiceEndpoint: EndpointProtocol, EndpointCommonProtoc
         }
     }
 
-    internal final class AppSVCBackingClass: Hashable, Sendable {
+    internal final class AppSVCBackingClass: Hashable, @unchecked Sendable {
         static func == (
             lhs: ApplicationServiceEndpoint.AppSVCBackingClass,
             rhs: ApplicationServiceEndpoint.AppSVCBackingClass
@@ -70,17 +66,9 @@ public struct ApplicationServiceEndpoint: EndpointProtocol, EndpointCommonProtoc
         func copy() -> Self {
             .init(storage: self.storage)
         }
-        var storage: Storage {
-            get {
-                _storage.withLock { $0 }
-            }
-            set {
-                _storage.withLock { $0 = newValue }
-            }
-        }
-        private let _storage: NetworkMutex<Storage>
+        var storage: Storage
         init(storage: Storage) {
-            self._storage = NetworkMutex(storage)
+            self.storage = storage
         }
     }
     typealias Backing = AppSVCBackingClass
--- a/Sources/SwiftNetwork/Endpoint/ApplicationServiceEndpoint.swift
+++ b/Sources/SwiftNetwork/Endpoint/ApplicationServiceEndpoint.swift
@@ -12,6 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(Synchronization)
+internal import Synchronization
+#endif
+
 @_spi(Essentials)
 @available(Network 0.1.0, *)
 public struct ApplicationServiceEndpoint: EndpointProtocol, EndpointCommonProtocol, Sendable {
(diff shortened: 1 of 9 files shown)


Sources/SwiftNetwork/Endpoint/ApplicationServiceEndpoint.swift
```diff
@@ -66,9 +66,22 @@ public struct ApplicationServiceEndpoint: EndpointProtocol, EndpointCommonProtoc
...
+				}
+			}
+			set {
+				lock.withLock { _ in
+					_storage = newValue
+				}
+			}
+		}
+		private nonisolated(unsafe) var _storage: Storage
+		private let lock = NetworkMutex(())
```
tfpauly: Would it work to put the mutex explicitly on the storage, like to make it `var storage = NetworkMutex<Storage>()`?

porchdog: Only reason I didn't do that was Storage doesn't have default initialization and it wasn't clear whether that was intentional or not, so I chose not to change that.

Sources/SwiftNetwork/Endpoint/BonjourEndpoint.swift
```diff
@@ -68,9 +68,22 @@ public struct BonjourEndpoint: EndpointProtocol, EndpointCommonProtocol {
...
+				}
+			}
+			set {
+				lock.withLock { _ in
+					_storage = newValue
+				}
+			}
+		}
+		private nonisolated(unsafe) var _storage: Storage
+		private let lock = NetworkMutex(())
```
agnosticdev: I think this may have been what Tommy was mentioning above but could we do:
```
    private final class BonjourBackingClass: Hashable, Sendable {
        static func == (lhs: BonjourEndpoint.BonjourBackingClass, rhs: BonjourEndpoint.BonjourBackingClass) -> Bool {
            lhs.storage == rhs.storage
        }
        public func hash(into hasher: inout Hasher) {
            hasher.combine(storage)
        }
        internal struct Storage: Hashable, Sendable {
            public var common: EndpointCommon
            public let name: String  // Name is informational for application-service endpoints.
            public let type: String
            public let domain: String
            public let composite: String
        }
        func copy() -> Self {
            .init(storage: self.storage)
        }
        var storage: Storage {
            get {
                _storage.withLock { $0 }
            }
            set {
                _storage.withLock { $0 = newValue }
            }
        }
        private let _storage: NetworkMutex<Storage>
        init(storage: Storage) {
            self._storage = NetworkMutex(storage)
        }
    }
```
And that way we do not have to use `nonisolated(unsafe)` ?

porchdog: Ah. Much better!

Sources/SwiftNetwork/Endpoint/ApplicationServiceEndpoint.swift
porchdog: Fixed in b50d76964d779624e50c9251c4b696417177f57a

Sources/SwiftNetwork/Endpoint/BonjourEndpoint.swift
porchdog: Fixed in b50d76964d779624e50c9251c4b696417177f57a

tfpauly (approved):

agnosticdev (approved):
Thank you!

Sources/SwiftNetwork/Endpoint/BonjourEndpoint.swift
rnro: I'm not very familiar with this code yet and maybe this is a performance improvement we can investigate down the line but I'm wondering if we can leverage `isKnownUniquelyReferenced` in some of these operations. This is commonly used in NIO, e.g. in the core of `ByteBuffer` https://github.com/apple/swift-nio/blob/48119dbbd23e3eabba48952ac7f75ebeeb87c217/Sources/NIOCore/ByteBuffer-core.swift#L1409

rnro: Ah, I see it _is_ used some places in here, I'll keep reading 😄

rnro:
The way that this pattern (a class-backed-struct) is made `Sendable` in the NIO family of repositories is to use the CoW as the concurrency protection by avoiding shared mutable state. The `struct` is then marked as `@unchecked Sendable` (not ideal but still required I think) e.g. https://github.com/apple/swift-nio/blob/48119dbbd23e3eabba48952ac7f75ebeeb87c217/Sources/NIOCore/ByteBuffer-core.swift#L1483 .

I suspect that we want to take that approach here since we are already using the NIO CoW pattern.

porchdog:
Add CoW in 67c1674499f5bd9f5c9ee2c592f2d6db5bcd67e3

PR #85, opened 2026-08-14
Make hashing cheaper on hot paths

glbrntt:
Motivation:

Hasher.init and Hasher.finalize show up heavily in QUIC benchmarks
(~4-5% of cycles for 1KB download for 100 concurrent streams). There are
three main sources for this in swift-network:

1. The transport parameters map,
2. The knownFlows map in QUICConnection, and
3. the MultiplexedFlowIdentifier lookup.

The transport parameters stores at most 19 elements and is keyed by an
enum: this alone is just shy of ~1% of the cycles. The known flows
lookup is also around ~1%, while the multiplexed flow ID accounts for
the rest.

Modifications:

- Back TransportParameters by an array of optionals instead of a
  dictionary: the transport parameters types are each assigned a fixed
  index in the array. This eliminates hashing there altogether at the
  cost of ever so slightly more memory.
- Key the known flows map in QUICConnection by the raw value for the
  stream ID instead of the nominal type. Swift special cases the hashing
  of various types to elide the creation of Hasher.
- Implement the '_rawHashValue' customisation point in
  MultiplexedFlowIdentifier: this also avoids creating a Hasher.
- The hash computation of MultiplexedFlowIdentifier also uses two bits
  as a discriminator for the underlying case.

Result:

The first two optimisation increase throughput in 1KB download QUIC
benchmarks by ~1.9%. Adding the multiplexed flow identifier change in
raises that to ~4.9%.


Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -166,16 +166,25 @@ where Path.LowerProtocol == OutboundDatagramLinkage {
...
+            return UInt64(bitPattern: Int64(index)) << 2 | 0b01
         case .inboundFlow(let index):
-            hasher.combine(index)
+            return UInt64(bitPattern: Int64(index)) << 2 | 0b10
         }
     }
 
+    public func _rawHashValue(seed: Int) -> Int {
+        self.rawHashKey._rawHashValue(seed: seed)
+    }
```
glbrntt, line 186: I would love to get @lorentey's opinion on the use of this particular API since it's underscored...

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -299,7 +299,10 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
     // List of streams that have app input data in their reassembly queue.
     var pendingReassemblyDequeue = QUICStreamList.pendingReassemblyDequeueList()
 
-    private(set) var knownFlows = [QUICStreamID: MultiplexedFlowIdentifier]()
+    // The logical key choice is 'QUICStreamID', however, Swift special cases the hashing of
```
tfpauly: @glbrntt what if we just made `QUICStreamID` be `RawRepresentable` and just be a raw `UInt64`? That's all it stores anyway.

glbrntt: I looked into that: sadly `RawRepresentable` hits the slow path! The default implementation of `_rawHashValue(seed:)` is this:

```
public func _rawHashValue(seed: Int) -> Int {
  var hasher = Hasher(_seed: seed)
  hasher.combine(self)
  return hasher._finalize()
}
```

So for `RawRepresentable` to get the fast path of not creating the `Hasher` it would have to forgo allowing `combine(_:)` to be customized.

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -299,7 +299,10 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
...
+    // various primitives (by avoiding the construction of a Hasher altogether). The result is
+    // that hashing the raw value is significantly cheaper which adds up on hot paths.
+    private(set) var knownFlows = [UInt64: MultiplexedFlowIdentifier]()
```
agnosticdev: Nice, thank you for uncovering this @glbrntt !

agnosticdev (approved):

Tests/SwiftNetworkTests/SwiftNetworkTransportParametersTests.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+@_spi(Essentials) @_spi(ProtocolProvider) @testable import SwiftNetwork
+
+@available(Network 0.1.0, *)
+final class SwiftNetworkTransportParametersTests: XCTestCase {
```
rpaulo: Shouldn't this test be part of the QUICTests instead?

Sources/SwiftNetwork/QUIC/QUICConnection.swift
rpaulo: We may want to consider typealias QUICStreamID = UInt64 and then deal with the methods as extensions to UInt64 to avoid more problems like this?

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
lorentey: This is a great use of `_rawHashValue`! Forwarding to UInt64's implementation still ensures strong hashing, and marking the case in the two lower bits helps reduce collisions in the original implementation. 👍

Sources/SwiftNetwork/QUIC/QUICConnection.swift
glbrntt: Given Karoy gave the thumbs up on using `_rawHashValue(seed:)` I've switched to using it on `QUICStreamID` too so we maintain the "right" API with the fast path.

Tests/SwiftNetworkTests/SwiftNetworkTransportParametersTests.swift
glbrntt: Yes – good catch!

rpaulo (approved):

PR #87, opened 2026-08-14
CI: extract build/test commands into shared scripts

rpaulo:
- Move macOS and Linux build/test commands out of the workflow file into
  dedicated `ci-macos.sh` and `ci-linux.sh` scripts backed by a shared
  `ci-support.sh` step runner.
- Run all steps even when one fails, print a consolidated summary (also
  to the run summary page), and exit non-zero if any step failed.
- Split `nightly-main` into its own job, currently skipped due to a
  compiler crash in the LoadableByAddress SIL pass.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -13,62 +13,28 @@ jobs:
       swift_nightly_flags: "--build-system native -Xswiftc -warnings-as-errors"  # workaround nightly-main issue
       enable_macos_checks: true
       macos_exclude_xcode_versions: "[{\"xcode_version\": \"swift_6.1\"}, {\"xcode_version\": \"swift_6.2\"}]"
-      macos_build_command: >-
-        bash -c '
-        rc=0;
-        echo "=== xcodebuild macOS ===";
-        /usr/bin/xcodebuild -quiet -scheme swift-network-evolution-Package -destination "generic/platform=macos,variant=macos" build && macos=PASSED || { rc=1; macos=FAILED; };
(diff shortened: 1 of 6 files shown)


.github/workflows/pull_request.yml
```diff
@@ -13,62 +13,28 @@ jobs:
...
+      enable_android_sdk_build: false
+      enable_embedded_wasm_sdk_build: false
+      enable_linux_static_sdk_build: true
+  tests-nightly-main:
+    name: Test (nightly-main)
+    # Skipped: the nightly-main (6.5-dev) compiler cannot build this package.
+    # It crashes in the LoadableByAddress SIL pass ("Unimplemented user") on
+    # Frame.span, so every build and test configuration fails. Flip this back to
+    # true once the upstream fix lands.
+    if: false
```
rpaulo, line 30: Looks like there's a fix coming soon for this, so we should be able to remove this line soon.

rpaulo:
@tfpauly @rnro @agnosticdev  please review this change.  Once we get a new swift-ci toolchain for nightly main, we can just remove the `if: false` line (and the comment).

This change accomplishes two things:
1. Makes the build scripts non-inline which are easier to read and change
2. Allows us to selectively disable nightly main if needed in the future.

Both are improvements IMO.  Since the resolution for the swift toolchain is coming soon, I think we can wait for it rather than build a weekly workflow like we discussed yesterday.

agnosticdev (approved):
I am fine here just please enable this once the nightly main is passing again.

rnro (approved):

PR #88, opened 2026-08-16
SwiftQUIC: PERF: Reduce CPU using in processIncomingStreamData by 29%

agnosticdev:
The stream function processIncomingStreamData is the per-packet hot-path for just about every workload we have.
There are two large sources of CPU usage in the function:

1. Processing items in the ReassemblyQueue
2. Exclusivity access from a class to a non-copyable struct (swift_beginAccess / swift_endAccess)

This change is designed to reduce exclusivity access by reducing the touch points to `FlowControl`, `QUICStreamList`, and `ReassemblyQueue`.
This change also removes some redundant code and trims down the arithmetic guards in `ReassemblyQueue`.
Overall this change results in a 29% reduction to CPU in the hot path.

Top of tree:
```
718.46 M  	 QUICStreamInstance.processIncomingStreamData(connection:frame:)	
212.03 M 	   ReassemblyQueue.append(frame:offset:fin:)	
15.00 M  	    specialized UnsafeMutablePointer<>.initialize(to:)	
5.00 M          ReassemblyQueueItem.length.getter	
129.00 M  	   QUICStreamInstance.updateInboundFlowControlCredit(dataLengthAdded:connection:connectionOnly:)	
89.58 M  	   swift_beginAccess	
64.84 M   	   <deduplicated_symbol>	
47.00 M   	   QUICStreamInstance.startTrackingInboundFlowControlInterval(connection:)	
42.00 M    	   swift_endAccess	
18.00 M  	   QUICStreamInstance.updateLastOffset(connection:newLastOffset:newFinalSize:)	
6.00 M    	   DYLD-STUB$$swift_beginAccess	
6.00 M         QUICStreamList.append(_:)	
3.00 M         QUICStreamInstance.updateLastReceivedOffset(to:connection:)	
2.00 M         DYLD-STUB$$swift_endAccess	
1.00 M         DYLD-STUB$$swift_retain	
```

With this change:
```
507.09 M 	QUICStreamInstance.processIncomingStreamData(connection:frame:)	
153.03 M 	 ReassemblyQueue.append(frame:offset:fin:)	
86.26 M  	 <deduplicated_symbol>	
58.00 M      QUICStreamInstance.updateInboundFlowControlCredit(dataLengthAdded:connection:connectionOnly:)	
56.01 M  	 <deduplicated_symbol>	
42.01 M  	 swift_beginAccess	
16.00 M 	 QUICStreamInstance.startTrackingInboundFlowControlInterval(connection:)	
9.00 M     	 QUICStreamInstance.updateLastOffset(connection:newLastOffset:newFinalSize:)	
7.00 M    	 swift_endAccess	
6.00 M     	 DYLD-STUB$$swift_endAccess	
3.00 M     	 DYLD-STUB$$swift_beginAccess	
3.00 M       QUICStreamList.append(_:)
```

### DIFF
--- a/Sources/SwiftNetwork/QUIC/FlowControl.swift
+++ b/Sources/SwiftNetwork/QUIC/FlowControl.swift
@@ -309,7 +309,7 @@ extension QUICConnection {
         )
     }
 
-    @inline(__always)
+    @inline(always)
     func sendInboundFlowControlCredit() {
         guard !state.isTerminal else {
             return
--- a/Sources/SwiftNetwork/QUIC/FlowControl.swift
+++ b/Sources/SwiftNetwork/QUIC/FlowControl.swift
@@ -200,6 +200,7 @@ struct FlowControlState: ~Copyable {
     }
 
     // Returns true if changed
+    @inline(always)
     fileprivate mutating func recalculateInboundMaxData() -> Bool {
         // The new inbound max data is equal the number of bytes already delivered
         // to the application, plus the number of bytes we are willing to add
@@ -308,6 +309,7 @@ extension QUICConnection {
         )
     }
 
+    @inline(__always)
     func sendInboundFlowControlCredit() {
         guard !state.isTerminal else {
             return
@@ -319,6 +321,7 @@ extension QUICConnection {
         }
     }
 
+    @inline(always)
     fileprivate func updateMaximumUnreadInboundBytesAllowed(increment: UInt64) {
         let oldValue = flowControlState.maximumUnreadInboundBytesAllowed
         flowControlState.maximumUnreadInboundBytesAllowed = min(
@@ -496,6 +499,7 @@ extension QUICStreamInstance {
         )
     }
 
+    @inline(always)
     fileprivate func sendInboundFlowControlCredit(
         connection: QUICConnection,
         sendStreamCredit: Bool,
@@ -603,6 +607,7 @@ extension QUICStreamInstance {
         return min(connectionCredit, streamCredit)
     }
 
+    @inline(always)
     func updateLastReceivedOffset(
         to newLastReceivedOffset: UInt64,
         connection: QUICConnection
@@ -638,6 +643,7 @@ extension QUICStreamInstance {
         return delta
     }
 
+    @inline(always)
     func startTrackingInboundFlowControlInterval(connection: QUICConnection) {
         // Start of a new measurement interval
         if flowControlStreamState.receiveHighWaterMarkTime == .zero {
@@ -647,6 +653,7 @@ extension QUICStreamInstance {
 
     // Updates inbound flow control credit, and recalculates
     // `maximumUnreadInboundBytesAllowed` values.
+    @inline(always)
     func updateInboundFlowControlCredit(
         dataLengthAdded: UInt64,
         connection: QUICConnection,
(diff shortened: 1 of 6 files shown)


Sources/SwiftNetwork/QUIC/FlowControl.swift
```diff
@@ -308,6 +309,7 @@ extension QUICConnection {
         )
     }
 
+    @inline(__always)
```
glbrntt: nit:

```suggestion
    @inline(always)
```

Sources/SwiftNetwork/QUIC/QUICStream.swift
```diff
@@ -328,7 +328,7 @@ struct StreamListMembership: OptionSet {
 // Also note that a flow identifier can exist in multiple lists at one time.
 @available(Network 0.1.0, *)
 struct QUICStreamList: ~Copyable {
-    private var list = Deque<MultiplexedFlowIdentifier>()
+    private var list = Deque<MultiplexedFlowIdentifier>(minimumCapacity: 1)
```
glbrntt: I get a bit nervous about this pattern... an empty init for `QUICStreamList` will allocate unconditionally but may never be used. 

Not familiar enough with where this is used but is it possible to init with no capacity and then from context work out approximately how many items may be appended and then reserve capacity? That'd also avoid any intermediate reallocs.

Sources/SwiftNetwork/QUIC/ReassemblyQueue.swift
```diff
@@ -91,13 +91,33 @@ struct ReassemblyQueue: ~Copyable {
         #endif
     }
 
-    // Append data to the reassembly queue. Returns the amount of data added.
+    struct AppendResult {
+        var sizeAdded: Int
+        var lastOffset: Int
+        var hasFin: Bool
```
glbrntt: probably makes no practical difference but `hasFin` will have 7 bytes of padding after it: move it to the end and `AppendResult` will get 7 bytes smaller.

agnosticdev: Yeah that's a good call. [5fdfc69](https://github.com/apple/swift-network-evolution/pull/88/commits/5fdfc69275eed4c3ed972896d66672f7ccb4d229)

Sources/SwiftNetwork/QUIC/FlowControl.swift
agnosticdev: Addressed. [5fdfc69](https://github.com/apple/swift-network-evolution/pull/88/commits/5fdfc69275eed4c3ed972896d66672f7ccb4d229)

Sources/SwiftNetwork/QUIC/QUICStream.swift
agnosticdev: This should always be used if the list is created.  The only way it would not be used is a connection that fails without a stream or a stream that has no frames on it.

glbrntt: Okay, fair enough. Is there any value in a slightly larger initial capacity? e.g. 4 or so? That would avoid a few intermediate reallocs but is somewhat dependent on what the common case is.

agnosticdev: This is a good point [b8d3d3a](https://github.com/apple/swift-network-evolution/pull/88/commits/b8d3d3aff285f28d9ccbc2f3722c00a950f94b3b)

agnosticdev: Raised it to 4 in the latest

glbrntt (approved):

tfpauly (approved):

PR #89, opened 2026-08-17
SwiftQUIC: PERF: Delivery data directly to the flow

agnosticdev:
This is a minor change for overloading `deliverInboundDatagrams` and `deliverInboundStreamData` so the stack does not have to do a flow lookup twice.  Saves a minimal amount of work but in the hot path this will reduce instructions.

### DIFF
--- a/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
@@ -1400,18 +1400,16 @@ extension ManyToManyApplicationStreamProtocol where Flow: AutomaticUpperStreamPr
         streamData: consuming FrameArray
     ) throws(NetworkError) {
         guard var flow = self.flow(for: flowID) else { throw NetworkError.posix(EINVAL) }
-        try flow.addToUpperReceiveQueue(streamData)
-        flow.serviceUpperReceiveQueue()
+        try deliverInboundStreamData(flow: &flow, streamData: streamData)
     }
 
     // Enqueue and deliver the stream data directly to the flow
     public func deliverInboundStreamData(
-        flow existingFlow: Flow?,
+        flow existingFlow: inout Flow,
         streamData: consuming FrameArray
     ) throws(NetworkError) {
-        guard var flow = existingFlow else { throw NetworkError.posix(EINVAL) }
-        try flow.addToUpperReceiveQueue(streamData)
-        flow.serviceUpperReceiveQueue()
+        try existingFlow.addToUpperReceiveQueue(streamData)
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -1403,6 +1403,16 @@ extension ManyToManyApplicationStreamProtocol where Flow: AutomaticUpperStreamPr
         try flow.addToUpperReceiveQueue(streamData)
         flow.serviceUpperReceiveQueue()
     }
+
+    // Enqueue and deliver the stream data directly to the flow
+    public func deliverInboundStreamData(
+        flow existingFlow: Flow?,
```
tfpauly: Why is this flow optional?

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
tfpauly, line 1398: Let's have this call into the new `deliverInboundStreamData` function

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -1648,6 +1658,16 @@ extension HeterogeneousManyToManyProtocolHandler where SecondaryFlow: AutomaticU
         try flow.addToUpperReceiveQueue(datagrams)
         flow.serviceUpperReceiveQueue()
     }
+
+    // Enqueue and deliver the datagrams directly to the flow
+    public func deliverInboundDatagrams(
```
tfpauly, line 1660: Same comments here

agnosticdev: Addressed in [c898c59](https://github.com/apple/swift-network-evolution/pull/89/commits/c898c59622b3f571f76aa61cb40218b942028fb2)

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
agnosticdev: Addressed in [c898c59](https://github.com/apple/swift-network-evolution/pull/89/commits/c898c59622b3f571f76aa61cb40218b942028fb2)

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
agnosticdev: Addressed in [c898c59](https://github.com/apple/swift-network-evolution/pull/89/commits/c898c59622b3f571f76aa61cb40218b942028fb2)

tfpauly (approved):

PR #91, opened 2026-08-17
Add cross module optimization for benchmark targets

agnosticdev:
Adds `-cross-module-optimization` to our benchmark executables for cross module optimizations into SwiftNetwork.


rnro (approved):
LGTM. It might be worth factoring this out and/or delineating benchmark targets with a comment but it's your call.

agnosticdev:
Closing as this is something that may skew our results.

PR #93, opened 2026-08-17
SwiftQUIC: PERF: Avoid calling inboundStarting and inboundStopped when not applicable

agnosticdev:
This change avoids calling `inboundStarting` and `inboundStopped` for each key state every time `serviceReceivedDatagrams` and `inboundStopping` are called.  These only accumulated to around 6 megacycles in the QUICTransfer benchmark but now they are completely optimized away.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/QUICConnection.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICConnection.swift
@@ -1561,9 +1561,7 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
         // Detect if any received packet contains a QUIC Frame that unblocks
         // all streams, such as a new MAX_DATA. Includes setting
         // triggerAllStreamsUnblocked = false
-        initialPendingItems.inboundStarting()
-        handshakePendingItems.inboundStarting()
-        applicationPendingItems.inboundStarting()
+        withPendingItemsForKeyState { $0.inboundStarting() }
 
         // Tell recovery that a batch of packets is starting to be processed; suppress timer updates.
         // Ending recovery is deferred until servicing is done.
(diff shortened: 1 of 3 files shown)


Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -1561,9 +1561,7 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
         // Detect if any received packet contains a QUIC Frame that unblocks
         // all streams, such as a new MAX_DATA. Includes setting
         // triggerAllStreamsUnblocked = false
-        initialPendingItems.inboundStarting()
-        handshakePendingItems.inboundStarting()
-        applicationPendingItems.inboundStarting()
+        withPendingItemsForKeyState { $0.inboundStarting() }
```
tfpauly: I don't see how this is safe — when we are processing inbound packets, we don't know that the packets will belong to one list or another. They can be out of order.

However, overall, I question the usefulness of inboundStarting and inboundStopped being their own functions.

InboundStarting just sets triggerAllStreamsUnblocked, which is only used on application layers.

inboundStopped just does a precondition.

I would suggest we remove inboundStopped and we replace InboundStarting with a single call we do to just application state to mark triggerAllStreamsUnblocked = false.

tfpauly (changes requested):

Sources/SwiftNetwork/QUIC/QUICConnection.swift
agnosticdev: Good point, addressed in [e8ad159](https://github.com/apple/swift-network-evolution/pull/93/commits/e8ad159430d860d95775246294e62d541e9d8694)

tfpauly (approved):

PR #97, opened 2026-08-18
Allow starting standalone timers on NetworkContext

tfpauly:
Previously, timers used the same namespace as protocol event indices (in order to always have a de-duped value available). This prevents timers from being used for other non-protocol reasons on the context.

This change:
- Makes the timer reference a locked, incrementing UInt64
- Makes protocols that allow scheduling store a timer reference
- Adds conveniences to NetworkContext to let timers be directly scheduled using a duration, and returning a reference that can be used to unschedule later

### DIFF
--- a/Sources/SwiftNetwork/Protocols/ProtocolEventManager.swift
+++ b/Sources/SwiftNetwork/Protocols/ProtocolEventManager.swift
@@ -785,10 +785,17 @@ extension ProtocolInstanceReference {
         }
     }
 
-    func scheduleWakeup(milliseconds: UInt64,
-                               timerReference: TimerReference) {
+    func scheduleWakeup(
+        milliseconds: UInt64,
+        timerReference: TimerReference
+    ) {
         let protocolEventStateIndex = protocolEventStateIndex()!
-        context.scheduleWakeup(index: protocolEventStateIndex, timerReference: timerReference, referenceToWakeup: self, milliseconds: milliseconds)
+        context.scheduleWakeup(
+            index: protocolEventStateIndex,
+            timerReference: timerReference,
+            referenceToWakeup: self,
+            milliseconds: milliseconds
+        )
     }
 
     func unscheduleWakeup(timerReference: TimerReference) {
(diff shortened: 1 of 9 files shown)


Sources/SwiftNetwork/Protocols/ProtocolEventManager.swift
```diff
@@ -784,16 +785,13 @@ extension ProtocolInstanceReference {
         }
     }
 
-    var timerReference: TimerReference {
-        TimerReference(index: self._protocolEventStateIndex?.rawValue ?? 0)
-    }
-
-    public func scheduleWakeup(milliseconds: UInt64) {
+    func scheduleWakeup(milliseconds: UInt64,
```
rpaulo: I think we should start thinking about converting `milliseconds` to `NetworkDuration`.  If we don't support sub-millisecond timers, we could explain that in the code as a comment and a precondition.

Sources/SwiftNetwork/QUIC/Timer.swift
```diff
@@ -143,7 +146,7 @@ final class Timer: PrefixedLoggable {
             log.debug("Stopping timer")
             timerCancelled = true
             wakeup = .idle
-            reference?.unscheduleWakeup()
+            reference?.unscheduleWakeup(timerReference: timerReference)
```
rpaulo, line 149: Probably we should rename `reference` to something less generic in the future.

rpaulo (approved):

rpaulo:
Would this allow us to write unit tests for NetworkContext timers?

Sources/SwiftNetwork/Protocols/ProtocolEventManager.swift
tfpauly: Yes, I agree. That's why I was changing the convenience API to use a duration!

Sources/SwiftNetwork/QUIC/Timer.swift
tfpauly: Yeah, I'll be moving this soon to be more on linkage/instances, and not reference

tfpauly:
> Would this allow us to write unit tests for NetworkContext timers?

Yeah, I added a couple new unit tests here, but we can certain do more! Any particular surfaces you're thinking of?

PayasR (approved):

agnosticdev (approved):

kpodosin (approved):

PR #100, opened 2026-08-19
Recompute maximumStreamDataSize when the stream send buffer drains

kkuk24:
The per-stream send-queue permit (maximumStreamDataSize) is latched to 0 when pendingOutboundBytesToSend exceeds the 10*MSS watermark, and only recomputed on an inbound MAX_STREAM_DATA. Recompute it on drain too, so getOutboundStreamDataRoomAvailable reopens and a parked writer resumes without waiting on the peer

### DIFF
--- a/Sources/SwiftNetwork/QUIC/FlowControl.swift
+++ b/Sources/SwiftNetwork/QUIC/FlowControl.swift
@@ -395,6 +395,14 @@ extension QUICStreamInstance {
         flowControlState.totalOutboundBytesSent += bytes
         connection.flowControlState.pendingOutboundBytesToSend -= bytes
         connection.flowControlState.totalOutboundBytesSent += bytes
+
+        // Draining can reopen a permit that transient backpressure latched to 0
+        if self.maximumStreamDataSize == 0 {
+            updateOutboundFlowControlCredit(connection: connection)
+            if self.maximumStreamDataSize > 0 {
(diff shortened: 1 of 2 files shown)


rpaulo (approved):

Sources/SwiftNetwork/QUIC/FlowControl.swift
```diff
@@ -398,6 +395,14 @@ extension QUICStreamInstance {
         flowControlState.totalOutboundBytesSent += bytes
         connection.flowControlState.pendingOutboundBytesToSend -= bytes
         connection.flowControlState.totalOutboundBytesSent += bytes
+
+        // Draining can reopen a permit that transient backpressure latched to 0
+        if self.maximumStreamDataSize == 0 {
+            updateOutboundFlowControlCredit(connection: connection)
+            if self.maximumStreamDataSize > 0 {
+                upper.deliverOutboundRoomAvailableEvent(reference)
```
rpaulo, line 403: I'm wondering if this is the right place to deliver the event.  Should we do it later ?

agnosticdev (approved):

Sources/SwiftNetwork/QUIC/FlowControl.swift
kkuk24: I will look for a better place for this notification. I agree that it ideally should not be per STREAM frame sent. Will take that up as an optimization

PR #101, opened 2026-08-19
Workflows: enable shell_check.

rpaulo:
Enabling shell_check as I don't see a reason why it's disabled.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -78,6 +78,6 @@ jobs:
     uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
     with:
       api_breakage_check_enabled: false
-      shell_check_enabled: false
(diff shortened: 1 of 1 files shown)


tfpauly:
@rpaulo only reason it was disabled was that it had started failing for infrastructure reasons in the past. Happy to try again.

tfpauly (approved):

.github/workflows/pull_request.yml
```diff
@@ -78,6 +78,6 @@ jobs:
     uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
     with:
       api_breakage_check_enabled: false
-      shell_check_enabled: false
+      shell_check_enabled: true
```
rnro:
```suggestion
```
I think you can just drop that line, it’ll default to true

PR #103, opened 2026-08-19
Workflows: add docs_check_macos_enabled=true

rpaulo:


rpaulo:
Closing because what we need is .spi.yml to enable documentation checks.  The macOS docs check target doesn't need to be enabled.

PR #106, opened 2026-08-21
SwiftQUIC: PERF: Make knownFlows lookup cheaper

agnosticdev:
In `QUICConnection` we do a streamID -> flowID -> QUICStream look for each inbound stream packet and this change optimizes the key hashing for QUICStreamID to be a UInt64. This cuts the lookup CPU usage down here by 44%.

Top of tree:
```
1.39 G   QUICConnection.processStreamFrame(_:)	
172.67   specialized Dictionary.subscript.getter	
```
With this change:
```
1.31 G   QUICConnection.processStreamFrame(_:)	
77.02 M  specialized Dictionary<>.subscript.getter	
```

### DIFF
--- a/Sources/SwiftNetwork/QUIC/QUICConnection.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICConnection.swift
@@ -300,7 +300,7 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
     // List of streams that have app input data in their reassembly queue.
     var pendingReassemblyDequeue = QUICStreamList.pendingReassemblyDequeueList()
 
-    private(set) var knownFlows = [QUICStreamID: MultiplexedFlowIdentifier]()
+    private(set) var knownFlows = [UInt64: MultiplexedFlowIdentifier]()
 
     private(set) var localCIDLength: Int = 0
     private var initialSourceConnectionID: QUICConnectionID?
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -300,7 +300,7 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
     // List of streams that have app input data in their reassembly queue.
     var pendingReassemblyDequeue = QUICStreamList.pendingReassemblyDequeueList()
 
-    private(set) var knownFlows = [QUICStreamID: MultiplexedFlowIdentifier]()
+    private(set) var knownFlows = [UInt64: MultiplexedFlowIdentifier]()
```
rpaulo, line 303: I thought we took care of this with @glbrntt 's change?

agnosticdev: Right, he does have a `_rawHashValue` function on QUICStreamID, that should also do the trick here. I thought his change only addressed the `MultiplexedFlowIdentifier` key.

agnosticdev:
Lets see if https://github.com/apple/swift-network-evolution/pull/85 addresses this one and if not I can revisit.

PR #107, opened 2026-08-21
SwiftQUIC: Provide path events when QUICPath changes or is validated

agnosticdev:
There is a need for swift-nio-quic to observe events when `QUICPath` changes or becomes validated. This typically happens during connection migration. It is not appropriate to expose `QUICPath` directly so this change exposes a new type called `QUICPathInfo` that carries local and remote Endpoint information about the new datagram path underneath QUIC when it changes. When `QUICPath` changes or becomes validated QUIC checks with the lower protocol (UDP) to obtain the endpoints of the new path below it and sends a `QUICEvent` that can be observed at the connection level through an object such as NewStreamFlowHarness and the application level can update needed info based on these events.

This is essentially the same code as this PR just rebase:
https://github.com/apple/swift-network-evolution/pull/55

### DIFF
--- a/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
@@ -265,12 +265,9 @@ where LowerProtocol: OutboundDataLinkage, ParentProtocol: ManyToManyDatapathProt
 
 @_spi(ProtocolProvider)
 @available(Network 0.1.0, *)
-public struct DatagramPathEndpoints: Sendable, Equatable {
+public struct DatagramPathEndpoints: Sendable {
     let local: AddressEndpoint
     let remote: AddressEndpoint
-    public static func == (lhs: DatagramPathEndpoints, rhs: DatagramPathEndpoints) -> Bool {
-        lhs.local == rhs.local && lhs.remote == rhs.remote
-    }
 }
 
 // MARK: Implementations
--- a/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
@@ -233,7 +233,7 @@ extension MultiplexingPathIdentifier {
 @_spi(ProtocolProvider)
 @available(Network 0.1.0, *)
 @frozen public enum MultiplexingPathEvent: CustomStringConvertible {
-    case available
+    case available(Endpoint?, Endpoint?, PathProperties?, Parameters?)
     case unavailable
     case established
 
@@ -262,13 +262,6 @@ public protocol MultiplexingPath: UpperProtocolHandler {
 @available(Network 0.1.0, *)
 public protocol MultiplexingDatapathPath: MultiplexingPath
 where LowerProtocol: OutboundDataLinkage, ParentProtocol: ManyToManyDatapathProtocol {}
-
-@_spi(ProtocolProvider)
-@available(Network 0.1.0, *)
-public struct DatagramPathEndpoints: Sendable {
-    let local: AddressEndpoint
-    let remote: AddressEndpoint
-}
 
 // MARK: Implementations
 
@@ -372,7 +365,11 @@ extension ManyToManyProtocolHandler {
         try newPath.attachLowerProtocol(lowerProtocol, remote: remote, local: local, parameters: parameters, path: path)
         if multiplexingPaths.isEmpty { newPath.pathIsPrimary = true }
         multiplexingPaths[newPath.identifier] = newPath
-        handlePathChanged(path: newPath.identifier, event: .available, isPrimary: newPath.pathIsPrimary)
+        handlePathChanged(
+            path: newPath.identifier,
+            event: .available(local, remote, path, parameters),
+            isPrimary: newPath.pathIsPrimary
+        )
     }
     #endif
 
@@ -991,7 +988,11 @@ extension ManyToManyDatapathProtocol where Path.ParentProtocol == Self, Path: In
         if path?.hasMigrationInfo == true { newPath.pathHasMigrationInfo = true }
         multiplexingPaths[newPath.identifier] = newPath
         if !isFirstPath {
-            handlePathChanged(path: newPath.identifier, event: .available, isPrimary: newPath.pathIsPrimary)
+            handlePathChanged(
+                path: newPath.identifier,
+                event: .available(local, remote, path, parameters),
+                isPrimary: newPath.pathIsPrimary
+            )
         }
     }
 }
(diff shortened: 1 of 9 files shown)


Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -263,6 +263,16 @@ public protocol MultiplexingPath: UpperProtocolHandler {
 public protocol MultiplexingDatapathPath: MultiplexingPath
 where LowerProtocol: OutboundDataLinkage, ParentProtocol: ManyToManyDatapathProtocol {}
 
+@_spi(ProtocolProvider)
+@available(Network 0.1.0, *)
+public struct DatagramPathEndpoints: Sendable, Equatable {
+    let local: AddressEndpoint
+    let remote: AddressEndpoint
+    public static func == (lhs: DatagramPathEndpoints, rhs: DatagramPathEndpoints) -> Bool {
```
tfpauly: Do you need the manual equitability? That didn't seem needed in QUICPathInfo.

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -263,6 +263,16 @@ public protocol MultiplexingPath: UpperProtocolHandler {
 public protocol MultiplexingDatapathPath: MultiplexingPath
 where LowerProtocol: OutboundDataLinkage, ParentProtocol: ManyToManyDatapathProtocol {}
 
+@_spi(ProtocolProvider)
+@available(Network 0.1.0, *)
+public struct DatagramPathEndpoints: Sendable, Equatable {
```
tfpauly: I'm not sure this concept belongs in a general ManyToManyProtocol. Why not just have the events be quic specific and deliver QUICPathInfo?

agnosticdev: Having the events be QUIC specific seemed odd in the protocol layer. There has always been a separation of keeping QUIC types out of the protocol layer and I was trying to follow that as much as possible here. Also, the path being validated is a QUIC concept.

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
agnosticdev: This was an artifact of the previous code.  Removed [0c0573d](https://github.com/apple/swift-network-evolution/pull/107/commits/0c0573d7bd277dba2b0e08f53b063cd3c63eb71d)

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
```diff
@@ -243,6 +251,8 @@ public enum QUICEvent: DomainSpecificNetworkProtocolEvent {
     case maxStreamsLimitUnidirectionalUpdated(maximumStreams: Int)
     case earlyDataRejected
     case receivedRemoteTransportParameters(transportParameters: [UInt8])
+    case pathChanged(_ info: QUICPathInfo)
+    case pathValidated(_ info: QUICPathInfo)
```
tfpauly, line 255: Should also have a path-not-validated (unreachable)

Sources/SwiftNetwork/Protocols/ProtocolControlHandlers.swift
```diff
@@ -924,6 +924,18 @@ extension ProtocolInstanceReference {
         }
     }
 
+    public func getPathEndpoints(
```
tfpauly: We shouldn't be adding a general call for path endpoints on something as generic as the protocol control handler

tfpauly: Instead of this, we should be saving the endpoints/parameters/path or whatever we want from the call stack of `attachLowerDatagramProtocolForNewPath`

tfpauly: We can add the info (local, remote, parameters, path) as associated values to the `available` case on `MultiplexingPathEvent`

agnosticdev: Addressed [13cf797](https://github.com/apple/swift-network-evolution/pull/107/commits/13cf797bdd3c6466928e4ba82c74b27442845216)

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
agnosticdev: Addressed [13cf797](https://github.com/apple/swift-network-evolution/pull/107/commits/13cf797bdd3c6466928e4ba82c74b27442845216)

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -233,7 +233,7 @@ extension MultiplexingPathIdentifier {
 @_spi(ProtocolProvider)
 @available(Network 0.1.0, *)
 @frozen public enum MultiplexingPathEvent: CustomStringConvertible {
-    case available
+    case available(Endpoint?, Endpoint?, PathProperties?, Parameters?)
```
tfpauly: Let's label these associated values, please

agnosticdev: Good call: [1ad427f](https://github.com/apple/swift-network-evolution/pull/107/commits/1ad427f5016bb849afde8404a0016afd6dca2503)

tfpauly (approved):

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -233,7 +233,7 @@ extension MultiplexingPathIdentifier {
 @_spi(ProtocolProvider)
 @available(Network 0.1.0, *)
 @frozen public enum MultiplexingPathEvent: CustomStringConvertible {
-    case available
+    case available(local: Endpoint?, remote: Endpoint?, path: PathProperties?, parameters: Parameters?)
     case unavailable
     case established
```
rpaulo, line 239: It's possible that when we bring a path up, the addresses may have changed.

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
```diff
@@ -19,6 +19,14 @@ public struct NetworkEventDomain: Sendable, Hashable, CustomStringConvertible {
     public var description: String { domain }
 }
 
+@_spi(ProtocolProvider)
+@available(Network 0.1.0, *)
```
rpaulo, line 23: is 0.1.0 the correct macro here?

rpaulo (approved):

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
agnosticdev: It looks like this needs to change with our package file.

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
agnosticdev: I added a comment regarding this case.

PR #108, opened 2026-08-21
Socket test improvements

rnro:
This PR aims to make Socket tests faster and more stable, no source changes.

- Servers now bind port 0 and report `listeningPort`.
- Where the API doesn't expose the port a test helper is used to discover a free port and that is then used (`discoverFreeLoopbackPorts`). This is still racy, but avoids most collisions and issues when hanging tests have the port.
- `TCPClientHarness.init` reports a start failure instead of dropping it with `try?`, so a bind failure no longer surfaces later as an unexplained timeout.
- The refused-connection test asserts from its recorded state history rather than an inverted expectation, which XCTest can only satisfy by waiting out the whole timeout (locally 0.002s against 5.003s).
- The split-send server blocks on the peer's FIN instead of a one second sleep.
- The interface test which previously exhausted all FDs on the system now lowers its own `RLIMIT_NOFILE` instead of opening sockets until the kernel refuses. This was probably a cause of a reasonable amount of test flakiness and was slow. (locally 17s to under a millisecond).
- Three waits are deliberately kept, each now commented with rationale

### DIFF
--- a/Tests/SwiftNetworkTests/SwiftNetworkInterfaceTests.swift
+++ b/Tests/SwiftNetworkTests/SwiftNetworkInterfaceTests.swift
@@ -19,7 +19,6 @@ import XCTest
 import Glibc
 #elseif canImport(Musl)
 import Musl
-#elseif canImport(os)
 #endif
 
 @available(Network 0.1.0, *)
--- a/Tests/SwiftNetworkTests/SwiftNetworkInterfaceTests.swift
+++ b/Tests/SwiftNetworkTests/SwiftNetworkInterfaceTests.swift
@@ -17,12 +17,9 @@ import XCTest
 
 #if canImport(Glibc)
 import Glibc
-internal import Logging
 #elseif canImport(Musl)
 import Musl
-internal import Logging
 #elseif canImport(os)
-internal import os
 #endif
 
 @available(Network 0.1.0, *)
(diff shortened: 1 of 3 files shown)


Tests/SwiftNetworkTests/SwiftNetworkSocketTests.swift
```diff
@@ -221,21 +223,21 @@ final class SwiftNetworkSocketTests: NetTestCase {
     // MARK: - Edge case payloads (UDP)
 
     func testEmptyPayload() {
-        let harness = UDPLoopbackHarness(basePort: 10950)
+        let harness = UDPLoopbackHarness()
```
agnosticdev, line 226: This is nice, thank you!

agnosticdev (approved):

Tests/SwiftNetworkTests/SwiftNetworkInterfaceTests.swift
```diff
@@ -17,12 +17,9 @@ import XCTest
 
 #if canImport(Glibc)
 import Glibc
-internal import Logging
 #elseif canImport(Musl)
 import Musl
-internal import Logging
 #elseif canImport(os)
-internal import os
```
PayasR, line 25: Is this just an empty #elseif branch now?

PayasR (approved):

Tests/SwiftNetworkTests/SwiftNetworkInterfaceTests.swift
rnro: Good spot

tfpauly (approved):

PR #109, opened 2026-08-21
`Cubic` pacing rate tolerates a zero smoothed RTT

rnro:
`Cubic.updatePacerState` divides by the smoothed RTT in microseconds, so a small smoothed RTT is rounded to zero and traps.

`Prague.updatePacerState` already uses its initial estimate in the case of a zero smoothed RTT, so this change mirrors that approach.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/Cubic.swift
+++ b/Sources/SwiftNetwork/QUIC/Cubic.swift
@@ -227,7 +227,7 @@ struct Cubic: CongestionControlProtocol, CubicLikeProtocol {
             return
         }
 
-        // A same-host RTT can round to zero: `RTT.processNewSample` stores the sample as whole
+        // A short RTT can round to zero: `RTT.processNewSample` stores the sample as whole
         // microseconds, so an ack duration under 500ns becomes 0µs and dividing by it below would
         // trap. Fall back to the initial estimate.
         let smoothedRTTInMicroseconds =
--- a/Sources/SwiftNetwork/QUIC/Cubic.swift
+++ b/Sources/SwiftNetwork/QUIC/Cubic.swift
@@ -227,22 +227,20 @@ struct Cubic: CongestionControlProtocol, CubicLikeProtocol {
             return
         }
 
-        // Zero microseconds means either that no measurement has landed yet or that one rounded
-        // down from under half a microsecond, and dividing by it below would trap. Fall back to the
-        // initial estimate in both cases.
+        // A same-host RTT can round to zero: `RTT.processNewSample` stores the sample as whole
+        // microseconds, so an ack duration under 500ns becomes 0µs and dividing by it below would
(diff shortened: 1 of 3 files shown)


Sources/SwiftNetwork/QUIC/Prague.swift
```diff
@@ -312,7 +315,7 @@ struct Prague: CongestionControlProtocol, CubicLikeProtocol {
         }
 
         // Multiply by USEC_PER_SEC as sRTT is in microseconds
```
rpaulo:
```suggestion
        // Multiply by USEC_PER_SEC as smoothedRTTInMicroseconds is in microseconds
```

Sources/SwiftNetwork/QUIC/Cubic.swift
```diff
@@ -226,13 +226,20 @@ struct Cubic: CongestionControlProtocol, CubicLikeProtocol {
...
+        let smoothedRTTInMicroseconds =
+            smoothedRTT.microseconds == 0 ? pacingInitialRTT.microseconds : smoothedRTT.microseconds
+
         var rate = congestionWindow
+
         // Use 200% rate when in slow start
         if congestionWindow < slowStartThreshold {
             rate *= 2
         }
-        // Multiply by USEC_PER_SEC as srtt is in microseconds
```
rpaulo, line 234: Keep this comment?

Sources/SwiftNetwork/QUIC/Cubic.swift
```diff
@@ -226,13 +226,20 @@ struct Cubic: CongestionControlProtocol, CubicLikeProtocol {
         guard let path, path.pacer.enabled else {
             return
         }
+
+        // Zero microseconds means either that no measurement has landed yet or that one rounded
```
rpaulo: How can this happen when we are called after ackEnd? At that time I expect we should have an RTT estimate.

rnro: You're right, I don't think the "no measurement has landed yet" part can happen. I've updated the comment.

Sources/SwiftNetwork/QUIC/Cubic.swift
rnro: So I've taken a slightly different approach here. I think the underlying issue once I called it `smoothedRTTInMicroseconds` is that the units of `rate` were not clear. I've changed that and I think then the comment is not needed because the code is self-evident.

Sources/SwiftNetwork/QUIC/Prague.swift
rnro: https://github.com/apple/swift-network-evolution/pull/109#discussion_r3855705284

rpaulo (approved):

tfpauly (approved):

PR #114, opened 2026-08-24
SwiftQUIC: PERF: Reduce exclusive access checks by 1.1 gigacycles in the send path

agnosticdev:
Swift has something called exclusive access checks, its part of what makes Swift a memory safe language. These exclusive access checks happen when accessing properties of a class or other reference type to make sure something else is not accessing and mutating that property at the same time.  These exclusive access checks end up costing a very small amount of  CPU when a property of a reference type is used.  Well, in the data path (millions of times) these very small amounts of CPU add up to becoming a lot of CPU.  One thing you can do to avoid these costs and still keep your code safe is to hoist these properties up to the top of the call stack and pass them down through the stack.  This ends up satisfying the ownership rules and avoids redundant exclusive access checks. And that's what this change does, it rewrites the send path to reduce calling `QUICConnection` properties directly on the send path, and instead, passes these properties down through the send path.  This change reduces the CPU usage by over 1.1 gigacycles.

Top of tree:
```
4.05 G 100.0%	-	 swift_beginAccess
775.78 M 100.0%	-	 swift_endAccess
```

With this change:
```
3.16 G 99.4%	-	 swift_beginAccess	
493.56 M 99.8%	-	  swift_endAccess
```

### DIFF
--- a/Sources/SwiftNetwork/QUIC/QUICConnection.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICConnection.swift
@@ -3002,6 +3002,8 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
             delayedACK: delayedACK,
             sentPackets: &sentPackets,
             recovery: &recovery,
+            initialPendingItems: &initialPendingItems,
+            handshakePendingItems: &handshakePendingItems,
             applicationPendingItems: &applicationPendingItems
         )
     }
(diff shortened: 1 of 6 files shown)


Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -3598,13 +3806,12 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
             }
             stats.increment(.txBytes, by: totalPacketLength)
             stats.increment(.txPackets)
-            withCurrentPath { (path: borrowing QUICPath) -> Void in
-                path.pathStatistics[.txPackets] += 1
-                path.pathStatistics[.txBytes] += Int(totalPacketLength)
-            }
+            path.pathStatistics[.txPackets] += 1
```
rpaulo: Why don't path statistics work the same way as connection statistics? path.stats.increment(.txPackets)

agnosticdev: I think I forgot to change path stats when I initially overhauled this (before this pr).  Overhauled them in connection.  Addressed [9ba1c9f](https://github.com/apple/swift-network-evolution/pull/114/commits/9ba1c9fc5bf8436e36524a89daa0c870b09afc3a)

tfpauly (approved):

PR #115, opened 2026-08-24
Schedule timers with a `NetworkDuration` instead of whole milliseconds

rnro:
`NetworkContext.Scheduler.schedule` took a count of milliseconds, so every caller computing a deadline at nanosecond resolution had to truncate on the way in. A delay under a millisecond truncated to no delay at all, arming a wakeup that could not reach the deadline it was armed for: the handler ran early, recomputed the same delay and armed again, spinning until enough real time had passed. `Timer.recalculate`, the bridge's link delay and `scheduleTimer(duration:)` all fed it.

* `Scheduler.schedule` takes an `after` delay, and `FutureTime` carries a duration, so the truncation has nowhere left to happen.
* `scheduleWakeup` threads the duration through `TimerSchedulable` and `ProtocolInstanceReference` to the context, replacing four `UInt64(_.milliseconds)` conversions.
* `DefaultScheduler` builds its deadline from whole seconds plus a remainder, so neither part can overrun the `Int` that `DispatchTimeInterval` takes, which is 32 bits on 32-bit watchOS.
* A recording `Scheduler` lets `testContextTimerKeepsASubMillisecondDelay` assert the delay a caller asked for.

### DIFF
--- a/Sources/SwiftNetwork/Context/NetworkContext.swift
+++ b/Sources/SwiftNetwork/Context/NetworkContext.swift
@@ -368,10 +368,16 @@ extension NetworkContext {
         /// Schedules a task to run after a delay, using a reference.
         func schedule(_ task: @escaping (() -> Void), after delay: NetworkDuration, reference: TimerReference) {
             let nanoseconds = max(delay.nanoseconds, 0)
+            let seconds = nanoseconds / 1_000_000_000
+            // A sub-second delay is the common case and its nanosecond count fits an `Int` on every
+            // platform, so it costs one addition. Longer delays are split because
+            // `DispatchTimeInterval` takes an `Int`, which is 32 bits on 32-bit watchOS.
             let targetTime =
-                DispatchTime.now()
-                + .seconds(Int(clamping: nanoseconds / 1_000_000_000))
-                + .nanoseconds(Int(nanoseconds % 1_000_000_000))
+                seconds == 0
+                ? DispatchTime.now() + .nanoseconds(Int(nanoseconds))
+                : DispatchTime.now()
+                    + .seconds(Int(clamping: seconds))
+                    + .nanoseconds(Int(clamping: nanoseconds % 1_000_000_000))
             globals.timerList.insert(targetTime: targetTime, reference: reference, task: task)
         }
         /// Unschedules a task with a reference.
@@ -417,6 +423,10 @@ extension NetworkContext {
 
     enum FutureTime {
         case unschedule
+        /// A delay and the task to run once it elapses.
+        ///
+        /// A negative delay runs the task at the first opportunity. The delay resolves nanoseconds,
+        /// but how finely a scheduler can honour it is the scheduler's own limit.
         case after(NetworkDuration, () -> Void)
     }
 
--- a/Sources/SwiftNetwork/Context/NetworkContext.swift
+++ b/Sources/SwiftNetwork/Context/NetworkContext.swift
@@ -83,9 +83,7 @@ public final class NetworkContext: NetworkContextProtocol, @unchecked Sendable {
         /// Runs an immediate task. No assumptions are made about how the task is run.
         func runImmediate(_ task: @escaping (() -> Void))
         /// Schedules a task to run after a delay, using a reference.
-        ///
-        /// The `milliseconds` parameter specifies the delay before the task runs.
-        func schedule(_ task: @escaping (() -> Void), milliseconds: Int64, reference: TimerReference)
+        func schedule(_ task: @escaping (() -> Void), after delay: NetworkDuration, reference: TimerReference)
         /// Unschedules a task with a reference.
         func unschedule(reference: TimerReference)
         /// A Boolean value that indicates whether the current code is running in the scheduler.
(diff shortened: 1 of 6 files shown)


Sources/SwiftNetwork/Context/NetworkContext.swift
```diff
@@ -368,10 +366,12 @@ extension NetworkContext {
...
         /// Schedules a task to run after a delay, using a reference.
-        ///
-        /// The `milliseconds` parameter specifies the delay before the task runs.
-        func schedule(_ task: @escaping (() -> Void), milliseconds: Int64, reference: TimerReference) {
-            let targetTime = DispatchTime.now() + DispatchTimeInterval.milliseconds(Int(milliseconds))
+        func schedule(_ task: @escaping (() -> Void), after delay: NetworkDuration, reference: TimerReference) {
+            let nanoseconds = max(delay.nanoseconds, 0)
+            let targetTime =
+                DispatchTime.now()
+                + .seconds(Int(clamping: nanoseconds / 1_000_000_000))
```
tfpauly: Why does this split up the seconds and nanoseconds?

rnro: If we put it all in at nanoseconds then we'd easily overrun 32-bit `Int` max (I think it's ~2.7s). Splitting it up this way means we can express bigger timer durations.

rnro: 32-bit `Int` max milliseconds in the old code was probably long enough to not cause an issue in practice.

Sources/SwiftNetwork/Context/NetworkContext.swift
```diff
@@ -83,9 +83,7 @@ public final class NetworkContext: NetworkContextProtocol, @unchecked Sendable {
         /// Runs an immediate task. No assumptions are made about how the task is run.
         func runImmediate(_ task: @escaping (() -> Void))
         /// Schedules a task to run after a delay, using a reference.
-        ///
-        /// The `milliseconds` parameter specifies the delay before the task runs.
-        func schedule(_ task: @escaping (() -> Void), milliseconds: Int64, reference: TimerReference)
+        func schedule(_ task: @escaping (() -> Void), after delay: NetworkDuration, reference: TimerReference)
```
agnosticdev, line 86: I think you marked this with the corresponding label but this will force a change here:
https://github.com/apple/swift-nio-quic/blob/main/Sources/NIOQUIC/SwiftNetwork/QUICChannelEventLoop.swift#L60

Sources/SwiftNetwork/Context/NetworkContext.swift
```diff
@@ -435,8 +435,8 @@ extension NetworkContext {
         switch time {
         case .unschedule:
             scheduler.unschedule(reference: reference)
-        case .milliseconds(let milliseconds, let block):
-            scheduler.schedule(block, milliseconds: Int64(milliseconds), reference: reference)
+        case .after(let delay, let block):
+            scheduler.schedule(block, after: delay, reference: reference)
```
agnosticdev, line 449: I am guessing its a wash, but just to make sure can you profile QUICTransfer with CPU trace to make sure that since we are operating on an object now that we did not incur a spike in CPU?

Sources/SwiftNetwork/Context/NetworkContext.swift
rnro: Yes, I believe that Tommy either has or is planning API-breaking changes so getting them all in the same release can be good for adopters.

Sources/SwiftNetwork/Context/NetworkContext.swift
tfpauly: Ah I see this is because of going through `DispatchTimeInterval`. Makes sense now!

Sources/SwiftNetwork/Context/NetworkContext.swift
tfpauly: Yeah this will be breaking

tfpauly (approved):

Sources/SwiftNetwork/Context/NetworkContext.swift
```diff
@@ -368,10 +366,12 @@ extension NetworkContext {
...
         }
         /// Schedules a task to run after a delay, using a reference.
-        ///
-        /// The `milliseconds` parameter specifies the delay before the task runs.
-        func schedule(_ task: @escaping (() -> Void), milliseconds: Int64, reference: TimerReference) {
-            let targetTime = DispatchTime.now() + DispatchTimeInterval.milliseconds(Int(milliseconds))
+        func schedule(_ task: @escaping (() -> Void), after delay: NetworkDuration, reference: TimerReference) {
+            let nanoseconds = max(delay.nanoseconds, 0)
+            let targetTime =
+                DispatchTime.now()
```
rpaulo: Shouldn't we be using System.Time.now?

rpaulo (approved):

Sources/SwiftNetwork/Context/NetworkContext.swift
rnro: From what I can see the only difference is that the construction of the `targetTime` now creates two `DispatchTimeInterval`s, one for the seconds and one for the nanoseconds. We could have the code branch in the case that there are 0 seconds, but I think this difference is in the noise anyway.

Sources/SwiftNetwork/Context/NetworkContext.swift
rnro: Possibly, it looks fishy here because the deadlines use system time but that is a broader change than this PR. `DispatchTime.now()` was used in the existing code and `timerList.insert(targetTime: ` requires a `DispatchTime`.

agnosticdev (approved):

PR #116, opened 2026-08-24
Thread 'now' through the stack

rnro:
This PR's primary goal is to lay the groundwork for making a clock/time injectable: while datapath code reads `NetworkClock.Instant.now` directly, no test can control it. This threads the instant through from the caller instead, so a later change can supply  it, e.g. from a scheduler.

Whilst doing this work I discovered a couple of potential issues/improvements.

Sampling the clock wherever a time was needed meant one inbound batch read it repeatedly, and loss detection compared recorded deadlines against an instant sampled later than the one its timer fired at.

Changes:
* Congestion control and loss detection take `now` from their caller; `QUICPath` supplies the instant `QUICConnection` already caches for the batch.
* `QLog` events take the timestamp of the event rather than of the write.
* `Timer` closures receive the instant the timer fired instead of reading the clock again.

Two behavior changes:

* The frames of an inbound batch share one receive timestamp, so inter-arrival within a batch now reads zero. An `@autoclosure` keeps the read out of stacks that never ask for timestamps.
* `CubicTests.testCubicCongestionLimited` and its Prague analogue now seed from a fixed instant. From the system clock they had send times in their own future, so every loss opened a new recovery period instead of one per round.

A few `NetworkClock.Instant.now` reads remain; those are where a scheduler-supplied clock would hook in.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/Ack.swift
+++ b/Sources/SwiftNetwork/QUIC/Ack.swift
@@ -232,7 +232,11 @@ struct AckSpace: ~Copyable, PrefixedLoggable {
             delay = 0
         case .applicationData:
             if delay == 0 {
-                delay = UInt64(largestTimestamp.duration(to: now).microseconds) >> delayExponent
+                // ACK Delay is an unsigned varint (RFC 9000 Section 19.3), and `UInt64` traps on a
+                // negative `Int64`. Clamp rather than widen: a delay measured as negative has no
+                // representation on the wire other than zero.
+                let elapsed = max(.zero, largestTimestamp.duration(to: now))
+                delay = UInt64(elapsed.microseconds) >> delayExponent
             }
 
         }
@@ -513,13 +517,14 @@ struct Ack: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
         }
     }
 
-    mutating func timerFired(timeNow: NetworkClock.Instant) {
+    mutating func timerFired(at timeNow: NetworkClock.Instant) {
         log.datapath("delayed ACK timer fired")
         if let connection = connection {
             if sendPending(
                 isAckSet: connection.isAckSet,
                 setAckFrame: connection.scheduleAckFrame,
-                ecn: connection.ecn
+                ecn: connection.ecn,
+                now: timeNow
             ) {
                 connection.sendFrames(delayedACK: true)
 
@@ -622,7 +627,8 @@ struct Ack: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
         delayExponent: Int,
         isAckSet: (PacketNumberSpace) -> Bool,
         setAckFrame: (PacketNumberSpace, consuming QUICFrame, Bool) -> Void,
-        ecn: borrowing ECN
+        ecn: borrowing ECN,
+        now: NetworkClock.Instant
     ) -> Bool {
         var shouldSend = false
         for packetNumberSpace in PacketNumberSpace.allCases {
@@ -649,7 +655,7 @@ struct Ack: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
                             delayExponent: delayExponent,
                             setAckFrame: setAckFrame,
                             ecnCounter: ecnCounter,
-                            now: path.parentProtocol.now
+                            now: now
                         )
                         shouldSend = shouldSend || ackSize > 0
                         ackSpace.needsTransmission = false
@@ -673,17 +679,19 @@ struct Ack: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
         on path: QUICPath,
         isAckSet: (PacketNumberSpace) -> Bool,
         setAckFrame: (PacketNumberSpace, consuming QUICFrame, Bool) -> Void,
-        ecn: borrowing ECN
+        ecn: borrowing ECN,
+        now: NetworkClock.Instant
     ) -> Bool {
         let shouldSend = assemble(
             for: path,
             delayExponent: localDelayExponent,
             isAckSet: isAckSet,
             setAckFrame: setAckFrame,
-            ecn: ecn
+            ecn: ecn,
+            now: now
         )
         if shouldSend {
-            sent(path.parentProtocol.now)
+            sent(now)
         }
         log.datapath("\(shouldSend)")
         return shouldSend
@@ -692,7 +700,8 @@ struct Ack: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
     mutating func sendPending(
         isAckSet: (PacketNumberSpace) -> Bool,
         setAckFrame: (PacketNumberSpace, consuming QUICFrame, Bool) -> Void,
-        ecn: borrowing ECN
+        ecn: borrowing ECN,
+        now: NetworkClock.Instant
     ) -> Bool {
         guard let connection else {
             return false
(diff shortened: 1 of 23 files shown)


agnosticdev (commented):
Given that `NetworkClock.Instant` is a copyable struct, and that you are passing this through functions in the IP data path can you profile QUICTransfer with and without this change to make sure we are not adding extra CPU calls for `outlined copy of NetworkClock.Instant` and then `outlined destroy of NetworkClock.Instant`? We'll want to make sure of that before proceeding here.

Tests/QUICTests/TestClock.swift
rpaulo, line 1: Since this in the QUICTests directory, does the file name need "QUIC"?

Tests/QUICTests/TestClock.swift
```diff
@@ -0,0 +1,67 @@
...
+@available(Network 0.1.0, *)
+extension NetworkClock.Instant {
+    /// A fixed instant to seed tests that need one.
+    ///
+    /// The congestion controllers only order instants and measure durations between
+    /// them, so the absolute value is arbitrary, but it has to be *fixed*. Seeding from
+    /// the real clock makes every duration depend on how long the test itself took to run.
+    ///
+    /// Non-zero because much of the stack treats `.zero` as "unset".
+    static var testBase: NetworkClock.Instant {
```
rpaulo, line 32: How about we re-use the value from the other test?

```
final class SwiftNetworkManualClockTests: NetTestCase {
    private let base = NetworkClock.Instant(milliseconds: 1000)
```

Tests/QUICTests/QUICTestClock.swift
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// Test-only conveniences that supply a fixed time.
+///
+/// The production signatures deliberately require a time, so that datapath code cannot
+/// silently reach for the real clock. Most ACK tests are about packet-number bookkeeping
+/// and not about time at all, so restating `now:` several hundred times would be noise.
+/// The default here is a *fixed* instant, so those tests stay deterministic. A test
+/// whose premise is timing calls the full form and passes its own instant.
+@available(Network 0.1.0, *)
```
rpaulo: Shouldn't this be in AckTests?

rnro:
> Given that `NetworkClock.Instant` is a copyable struct, and that you are passing this through functions in the IP data path can you profile QUICTransfer with and without this change to make sure we are not adding extra CPU calls for `outlined copy of NetworkClock.Instant` and then `outlined destroy of NetworkClock.Instant`? We'll want to make sure of that before proceeding here.

I was pretty sure that the compiler would just optimize through such a trivial object but to check I attempted to solve this by inspecting the binary. I ran
`nm ./QUICTransfer | swift demangle | grep -i 'outlined.*NetworkClock.Instant'`
which found no symbols produced for any outlined calls for that object, so I don't think they can have the performance implication you're concerned about. (I also ran the same command but grepping for `outlined.*Network`) and saw lots of results so I think this is a genuine result.

agnosticdev:
> > Given that `NetworkClock.Instant` is a copyable struct, and that you are passing this through functions in the IP data path can you profile QUICTransfer with and without this change to make sure we are not adding extra CPU calls for `outlined copy of NetworkClock.Instant` and then `outlined destroy of NetworkClock.Instant`? We'll want to make sure of that before proceeding here.
> 
> I was pretty sure that the compiler would just optimize through such a trivial object but to check I attempted to solve this by inspecting the binary. I ran `nm ./QUICTransfer | swift demangle | grep -i 'outlined.*NetworkClock.Instant'` which found no symbols produced for any outlined calls for that object, so I don't think they can have the performance implication you're concerned about. (I also ran the same command but grepping for `outlined.*Network`) and saw lots of results so I think this is a genuine result.

What does it say when you run the benchmark though in Instruments?

rnro:
> > > Given that `NetworkClock.Instant` is a copyable struct, and that you are passing this through functions in the IP data path can you profile QUICTransfer with and without this change to make sure we are not adding extra CPU calls for `outlined copy of NetworkClock.Instant` and then `outlined destroy of NetworkClock.Instant`? We'll want to make sure of that before proceeding here.
> > 
> > 
> > I was pretty sure that the compiler would just optimize through such a trivial object but to check I attempted to solve this by inspecting the binary. I ran `nm ./QUICTransfer | swift demangle | grep -i 'outlined.*NetworkClock.Instant'` which found no symbols produced for any outlined calls for that object, so I don't think they can have the performance implication you're concerned about. (I also ran the same command but grepping for `outlined.*Network`) and saw lots of results so I think this is a genuine result.
> 
> What does it say when you run the benchmark though in Instruments?

I think any change if one exists is within the noise
```
Before change: 64.50 G, 64.13 G, 64.77 G
After change:  64.63 G, 64.40 G, 64.66 G
```

agnosticdev (approved):

Tests/QUICTests/TestClock.swift
rnro: If we did that we'd have to introduce a dependency across test targets or create a new common re-usable target and I'm not sure that's worth the benefit.

Tests/QUICTests/TestClock.swift
rnro: Good point, I've dropped the QUIC

Tests/QUICTests/TestClock.swift
rpaulo: Why not put it under NETWORK_INTERNAL_TESTS in NetworkClock?

rnro: I personally don't think it's good practice to put this in the `Source` tree. Having it accessible from production code pollutes the namespace and creates a footgun. I'm also hoping to be able to remove `NETWORK_INTERNAL_TESTS` in the future if we have a fully injectable timing mechanism.

tfpauly (approved):

Sources/SwiftNetwork/QUIC/Ack.swift
```diff
@@ -232,7 +232,11 @@ struct AckSpace: ~Copyable, PrefixedLoggable {
             delay = 0
         case .applicationData:
             if delay == 0 {
-                delay = UInt64(largestTimestamp.duration(to: now).microseconds) >> delayExponent
+                // ACK Delay is an unsigned varint (RFC 9000 Section 19.3), and `UInt64` traps on a
+                // negative `Int64`. Clamp rather than widen: a delay measured as negative has no
+                // representation on the wire other than zero.
+                let elapsed = max(.zero, largestTimestamp.duration(to: now))
```
rpaulo, line 238: Is this really needed? Are we assuming that duration(to:) would return a negative number?

Sources/SwiftNetwork/QUIC/Migration.swift
```diff
@@ -85,14 +84,17 @@ struct Migration: ~Copyable {
...
-            timerNow: connection.now
+            timerNow: now
         )
     }
 
-    func timerFired(connection: QUICConnection) {
+    func timerFired(at firedAt: NetworkClock.Instant, connection: QUICConnection) {
         connection.log.debug("Migration timer fired")
 
-        sendPendingChallenges(connection: connection)
```
rpaulo, line 95: Why are we not sending probes here anymore?

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -3325,6 +3326,10 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
                 shouldEndBurst = true
             } else if packetBurst >= Constants.packetBurstCount {
                 // The packet burst count has been reached, check the time
+                //
+                // Deliberately the live clock rather than `self.now`: under a batch `self.now` is
```
rpaulo, line 3330:
```suggestion
                // Deliberately use the live clock rather than `self.now`: under a batch `self.now` is
```

Sources/SwiftNetwork/QUIC/Migration.swift
rnro: I _think_ that is answered in the comment below? I think we will still send them.

Sources/SwiftNetwork/QUIC/Ack.swift
rnro: The intention was to be defensive in case it did for whatever reason to avoid the trap.

PR #119, opened 2026-08-27
Support reading with spans in completions

tfpauly:
Support endpoint flow reading where a RawSpan is returned instead of allocating new bytes

### DIFF
--- a/Sources/SwiftNetwork/EndpointFlow/EndpointFlow.swift
+++ b/Sources/SwiftNetwork/EndpointFlow/EndpointFlow.swift
@@ -86,7 +86,7 @@ final class EndpointFlow: CustomDebugStringConvertible {
     let context: NetworkContext
     let identifier: UInt64
     var writeRequests = NetworkUniqueDeque<WriteRequest>()
-    var readRequests = [ReadRequest]()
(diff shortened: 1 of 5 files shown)


Sources/SwiftNetwork/EndpointFlow/EndpointFlow.swift
```diff
@@ -233,37 +233,58 @@ final class EndpointFlow: CustomDebugStringConvertible {
...
+            while !self.readRequests.isEmpty {
+                if self.readRequests[0].expectsSpan {
+                    guard var frames = flow.readFrames(minimumBytes: self.readRequests[0].minimumBytes,
+                                                       maximumBytes: self.readRequests[0].maximumBytes) else {
                         flow.waitForInboundDataAvailable(completion: self.inputAvailable)
                         break
                     }
+                    let readRequest = self.readRequests.removeFirst()
+                    frames.iterateMutableFrames { frame in
+                        if let bytes = frame.bytes {
```
agnosticdev, line 251: Anything else we can do here at the moment besides going through this computed property?

agnosticdev (approved):

PR #120, opened 2026-08-27
SwiftNetwork: PERF: Optimize UDP checksum16

agnosticdev:
This is a small optimization for UDP `checksum16` .  This change removes roughly 25% of the checksum instructions but we only see a small win of around 14% in the QUICTransfer benchmark:
```
// Before
768.13 M 100.0%	-	 checksum16()

// After
680.36 M 100.0%	-	 checksum16()
```
Assembly look:
<https://godbolt.org/z/GzWP1ea75>

### DIFF
--- a/Sources/SwiftNetwork/Protocols/Checksum.swift
+++ b/Sources/SwiftNetwork/Protocols/Checksum.swift
@@ -142,26 +142,13 @@ extension Frame {
             )
             throw ChecksumError.invalidLength
         }
-        // Operate directly on the underlying storage
-        switch buffer {
-        case .bytes:
-            let lowerBound = startOffset + offset
-            return _bytes.span.withUnsafeBytes { raw in
-                let offsetBuffer = UnsafeRawBufferPointer(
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/Protocols/Checksum.swift
```diff
@@ -52,7 +52,8 @@ extension UInt32 {
     }
 }
 
-enum ChecksumError: Error {
+@available(Network 0.1.0, *)
```
rpaulo: Who needs to know how the checksum failed?

agnosticdev: Needed for inlining.

rpaulo: I see.  Do we want to expose it just because of inlining though?

agnosticdev: Looks like I never tried `usableFromInline` :-)

rpaulo (approved):

tfpauly (approved):

PR #121, opened 2026-08-28
Server Initial packets should not carry tokens

josephnoir:
RFC 9000 says in Section 17.2.2 regarding token length:

> "Initial packets sent by the server MUST set the Token Length field to 0; clients that receive an Initial packet with a non-zero Token Length field MUST either discard the packet or generate a connection error of type PROTOCOL_VIOLATION."

If I read the code correctly, the server might falsely include the issued retry token in the packet.


rpaulo:
Isn't `initialToken` only set when we receive a retry? Don't we drop a RETRY from the client?

agnosticdev:
Yeah, I think the issue here is that once the server receives the new initial packet and validates it in `retryTokenPresent` then it continues to hold onto it and so its send in the `Packet.build` function.  You could destroy the token at that point because the packet and the token have been validated.  This would probably make more sense.  We will also need to change the unit tests in `SwiftNetworkQUICRetryTokenTests` to only validate the client token.

josephnoir:
> Isn't initialToken only set when we receive a retry? Don't we drop a RETRY from the client?

The server uses it to store the token it handed out to verify the client's response (and then includes in it's own response).

> You could destroy the token at that point because the packet and the token have been validated. This would probably make more sense.

Yeah, that sounds reasonable. I'll update the PR.

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -1643,6 +1643,11 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
                 return
             }
 
+            if state == .retrySent {
+                // The token was validated and is no longer needed.
+                self.initialToken = nil
```
rpaulo, line 1648: Can you move this to line 1635?

josephnoir: I was hesitant to have it earlier because there seem to be two code paths that exit after the token validation but before the state transition. And if we drop the token, `packetParser.retryTokenPresent(...)` does not have the token anymore.

Are the following checks guaranteed to succeed in that case?

rpaulo: I guess it's ok the way it is.

rpaulo (approved):

agnosticdev (approved):
Thank you @josephnoir !

PR #123, opened 2026-08-31
SwiftQUIC: Expose Stateless Reset Token when CID is retired

agnosticdev:
This change exposes an associated stateless reset token when a connection id is retired.
This is used on the server for cleanup purposes and tracking.


josephnoir (approved):
LGTM, thanks!

Sources/SwiftNetwork/Protocols/HarnessProtocols.swift
```diff
@@ -815,13 +815,27 @@ public class NewFlowHarness<LinkageType: InboundFlowLinkage, HarnessType: UpperH
 
     public var newInboundCIDEventCount = 0
     public var newOutboundCIDEventCount = 0
+    public var retiredOutboundCIDEventCount = 0
```
rpaulo, line 818: I think the QUIC specific code should be in an extension.

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
```diff
@@ -242,8 +242,8 @@ public enum QUICEvent: DomainSpecificNetworkProtocolEvent {
     #if !NETWORK_NO_SWIFT_QUIC
     case newInboundConnectionID(_ connectionID: QUICConnectionID)
     case retiredInboundConnectionID(_ connectionID: QUICConnectionID)
-    case newOutboundConnectionID(_ connectionID: QUICConnectionID)
-    case retiredOutboundConnectionID(_ connectionID: QUICConnectionID)
+    case newOutboundConnectionID(_ connectionID: QUICConnectionID, statelessResetToken: QUICStatelessResetToken)
```
rpaulo: Same comment about being in an extension

Tests/SwiftNetworkTests/SwiftNetworkQUICCIDTests.swift
```diff
@@ -0,0 +1,126 @@
...
+internal import os
+#endif
+
+#if IMPORT_SWIFTTLS
+#if canImport(SwiftTLS)
+// Uses cross module API for testing
+@available(Network 0.1.0, *)
+final class SwiftNetworkQUICCIDTests: NetTestCase {
+
+    func testQUICRetiredOutboundConnectionIDIncludesStatelessResetToken() {
```
rpaulo, line 55: I would argue that these should be in the QUICTest directory instead?

Sources/SwiftNetwork/Protocols/HarnessProtocols.swift
agnosticdev: We need to keep stored properties in the actual type.

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
agnosticdev: Moved this into a QUIC specific file: [e2377c1](https://github.com/apple/swift-network-evolution/pull/123/commits/e2377c110c06a5b804de0444ff6bcd26922dabb5)

Tests/SwiftNetworkTests/SwiftNetworkQUICCIDTests.swift
agnosticdev: We could split out the QUIC related files here but I think we should move that to a new PR.

Sources/SwiftNetwork/Protocols/HarnessProtocols.swift
tfpauly: I think the new file of having harness protocols have a "QUIC" extension that includes normal events like connected seems quite odd. I'd rather have it back

Sources/SwiftNetwork/Protocols/NetworkEvents.swift
tfpauly: That moved the harness protocol, not the events. It would make more sense to have a NetworkEvents+QUIC.swift file

agnosticdev: Updated [b674917](https://github.com/apple/swift-network-evolution/pull/123/commits/b6749179a56d51586b88b335352f32e18f126899)

rpaulo (approved):

tfpauly (approved):

PR #125, opened 2026-08-31
NetworkClock: fix duration arithmetic and improve descriptions

rpaulo:
- Fix float scaling, division, and shift operators to avoid premature integer truncation and preserve precision
- Add hours/days factories, coarse-unit descriptions, fractional microseconds, and roundedUpMilliseconds helper
- Make Instant subtraction return a NetworkDuration and render Instants using fine (sub-minute) units
- Cache nanosecond-per-unit constants and expand test coverage

### DIFF
--- a/Sources/SwiftNetwork/QUIC/FlowControl.swift
+++ b/Sources/SwiftNetwork/QUIC/FlowControl.swift
@@ -788,7 +788,7 @@ extension QUICStreamInstance {
                     shift = 1
                 }
                 log.datapath(
-                    "Estimated BDP \(receiveHighWaterMarkCount)B, current RTT \(rtt)us, estimated bandwidth 8 * \(receiveHighWaterMarkCount) / \(rtt) Mbps"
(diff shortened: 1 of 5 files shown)


tfpauly (approved):

agnosticdev (approved):
I would recommend going through and adding `&` to all of the places where its safe to do so since these values are often used in the hot path.  It doesn't seem like much but it removes a surprising amount of instructions when you do so.

rnro (approved):
Looks good overall, I agree with Matt's comment about `&+`

rpaulo:
> I would recommend going through and adding `&` to all of the places where its safe to do so since these values are often used in the hot path. It doesn't seem like much but it removes a surprising amount of instructions when you do so.

I don't think we should do that right now.  NetworkClock is a general-purpose subsystem and even if the current callers will never overflow or underflow, I don't think it's safe to assume that won't happen in the future.

PR #126, opened 2026-09-01
Count leading zeroes a few bytes at a time

glbrntt:
`QUICFrame.countZeroes()` shows up in a few benchmarks as using ~1.5% of CPU. It can be made a little faster by loading and checking 8 bytes at a time. For long runs of zeros (1200) it's ~4.5x faster, for very short runs (< 8) it's no worse.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/QUICFrame.swift
+++ b/Sources/SwiftNetwork/QUIC/QUICFrame.swift
@@ -736,21 +736,36 @@ struct FramePadding: ~Copyable, QUICFrameProtocol {
     private mutating func countZeros(frame: inout Frame) throws(QUICError) {
         var extraPadding = 0
 
-        if let bytes = frame.span {
-            let count = bytes.count
-            for byte in 0..<count {
-                if bytes[byte] == 0x00 {
-                    extraPadding += 1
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/QUIC/QUICFrame.swift
```diff
@@ -753,6 +746,28 @@ struct FramePadding: ~Copyable, QUICFrameProtocol {
...
+            if word == 0 {
+                offset &+= wordSize
+            } else {
+                return offset &+ (UInt64(littleEndian: word).trailingZeroBitCount / 8)
+            }
+        }
+
+        while offset < count, bytes[offset] == 0x00 {
+            offset &+= 1
+        }
```
agnosticdev, line 766: Is this for trailing bytes?

glbrntt: Yes: the first `while` loop does chunks of 8 bytes, this does the final up-to 7 seven bytes.

tfpauly (approved):

agnosticdev (approved):
Thank you @glbrntt !

PR #127, opened 2026-09-02
Guard and protect against faults/crashes on invalid requests

holleman:
- Make sure to not cause a fault when asking for a VLE to be put into too small of a type
- Make sure when deserializing uint8/bytes that a request does not exceed the bytes that are in the span or fast fail
- Add unit tests for these two conditions

### DIFF
--- a/Sources/SwiftNetwork/Utilities/Deserializer.swift
+++ b/Sources/SwiftNetwork/Utilities/Deserializer.swift
@@ -617,12 +617,12 @@ public struct Deserializer<Factory: DeserializerSpanFactory & ~Copyable & ~Escap
             return
         }
 
-        // Fast fail if total available bytes across all spans is insufficient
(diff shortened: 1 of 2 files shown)


Sources/SwiftNetwork/Utilities/Deserializer.swift
```diff
@@ -614,6 +617,11 @@ public struct Deserializer<Factory: DeserializerSpanFactory & ~Copyable & ~Escap
             return
         }
 
+        // Fast fail if total available bytes across all spans is insufficient
+        guard availableByteCount - totalBytesParsed >= byteCount else {
```
tfpauly: I'm concerned about perf impact of this check. I think it would be better to move this into the guard hasRoom check?

tfpauly (approved):

PR #129, opened 2026-09-03
SwiftQUIC: PERF: Avoid temporary frame array in the send path

agnosticdev:
This one is a double performance win, we save around 150 megacycles and we avoid a temporary allocation for FrameArray each time SwiftQUIC sends.
Today, each time SwiftQUIC sends frames it builds a temporary outboundFrameArray to add all of these frames in and then adds these frames to the lowerSendQueue.
We can just add these frames directly to the lowerSendQueue and take a CPU win too as long as we keep the capacity of the lowerSendQueue.

Top of tree:
```
29.97 G  88.8%	4.50 M  	  QUICConnection.buildSinglePacketForKeyState()	
29.87 G  88.5%	240.58 M	  specialized QUICConnection.buildSinglePacketForKeyState()	
220.14 M 0.7%	35.41 M 	  FrameArray.add(frame:)	


33.03 G 42.6%	8.00 M 	      QUICConnection.sendApplicationFrames()	
31.13 G 40.1%	44.91 M	       QUICConnection.runApplicationBurstLoop()	
1.55 G  2.0%	1.00 M 	       QUICConnection.sendOutboundFrames(_:on:)	
```
With this change:
```

29.49 G  88.5%	4.85 M  	   QUICConnection.buildSinglePacketForKeyState()	
29.44 G  88.4%	213.60 M	   specialized QUICConnection.buildSinglePacketForKeyState()	
172.28 M 0.5%	6.00 M  	   specialized ManyToManyOutboundDatagramProtocol<>.enqueueOutboundFrame(frame:path:)	

33.68 G 43.9%	1.00 M 	       QUICConnection.sendFrames()	
1.44 G  1.9%	-      	        QUICConnection.sendOutboundFrames(on:)	
```

### DIFF
--- a/Sources/SwiftNetwork/Protocols/FrameArray.swift
+++ b/Sources/SwiftNetwork/Protocols/FrameArray.swift
@@ -143,6 +143,14 @@ public struct FrameArray: ~Copyable {
         }
     }
 
+    // Make sure to keep the original capacity of the array to prevent reallocations
+    public mutating func drainArrayKeepingCapacity() -> FrameArray {
+        let count = self.count
+        let returnArray = self
+        self = FrameArray(capacity: count)
+        return returnArray
+    }
+
     public mutating func drainArray(maximumFrameCount: Int? = nil) -> FrameArray {
         if let maximumFrameCount, self.count > maximumFrameCount {
             var returnArray = FrameArray()
--- a/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
@@ -2059,9 +2059,9 @@ extension ManyToManyOutboundDatagramProtocol where Path: AutomaticLowerDatagramP
     }
 
     @inline(always)
-    public func enqueueOutboundFrame(frame: consuming Frame, path: Path) throws(NetworkError) {
+    public func enqueueOutboundFrame(frame: consuming Frame, path: Path) {
         var addPath = path
-        return try addPath.addToLowerSendQueue(frame)
+        return addPath.addToLowerSendQueue(frame)
     }
 
     @inline(always)
--- a/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
@@ -2058,18 +2058,6 @@ extension ManyToManyOutboundDatagramProtocol where Path: AutomaticLowerDatagramP
         return try path.addToLowerSendQueue(datagrams)
     }
 
-    @inline(always)
-    public func enqueueOutboundFrame(frame: consuming Frame, path: Path) {
-        var addPath = path
-        return addPath.addToLowerSendQueue(frame)
-    }
-
-    @inline(always)
-    public func sendEnqueuedOutboundDatagrams(path: Path) throws(NetworkError) {
-        var sendPath = path
-        sendPath.serviceLowerSendQueue()
-    }
-
     public func sendEnqueuedOutboundDatagrams(path pathID: MultiplexingPathIdentifier) throws(NetworkError) {
         guard var path = self.path(for: pathID) else { throw NetworkError.posix(EINVAL) }
         path.serviceLowerSendQueue()
--- a/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
+++ b/Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
@@ -2058,6 +2058,18 @@ extension ManyToManyOutboundDatagramProtocol where Path: AutomaticLowerDatagramP
         return try path.addToLowerSendQueue(datagrams)
     }
 
+    @inline(always)
+    public func enqueueOutboundFrame(frame: consuming Frame, path: Path) throws(NetworkError) {
+        var addPath = path
+        return try addPath.addToLowerSendQueue(frame)
+    }
+
+    @inline(always)
+    public func sendEnqueuedOutboundDatagrams(path: Path) throws(NetworkError) {
+        var sendPath = path
+        sendPath.serviceLowerSendQueue()
+    }
+
     public func sendEnqueuedOutboundDatagrams(path pathID: MultiplexingPathIdentifier) throws(NetworkError) {
         guard var path = self.path(for: pathID) else { throw NetworkError.posix(EINVAL) }
         path.serviceLowerSendQueue()
@@ -2116,7 +2128,7 @@ open class MultiplexingDatagramPath<ParentProtocol: ManyToManyOutboundDatagramPr
 
     public let identifier = MultiplexingPathIdentifier()
 
-    public var lowerSendQueue = FrameArray()
+    public var lowerSendQueue = FrameArray(capacity: 10)
     public var lowerReceiveQueue = FrameArray()
 
     public var pathIsPrimary: Bool = false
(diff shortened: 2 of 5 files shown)


Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
```diff
@@ -2058,6 +2058,18 @@ extension ManyToManyOutboundDatagramProtocol where Path: AutomaticLowerDatagramP
         return try path.addToLowerSendQueue(datagrams)
     }
 
+    @inline(always)
+    public func enqueueOutboundFrame(frame: consuming Frame, path: Path) {
```
tfpauly: If this function has a `path` already, the function should be on the path object, not on the connection. Same with the one below. 

Also, the aspects of having the pathID or path be used is a bit separate from using a frame array or not. It would be good to clean up any unused case

Sources/SwiftNetwork/Protocols/ProtocolDatagramHandlers.swift
```diff
@@ -57,12 +57,16 @@ extension AutomaticLowerDatagramProcessing where Self: ~Copyable {
         lowerSendQueue.add(frames: datagrams)
     }
 
+    public mutating func addToLowerSendQueue(_ datagram: consuming Frame) {
```
tfpauly, line 60: The same should be added to the upper datagram processing, upper stream processing, and lower stream processing

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -3230,14 +3230,9 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
         return outboundBatch
     }
 
-    private func sendOutboundFrames(_ outboundFrames: consuming FrameArray, on path: QUICPath) {
-        guard !outboundFrames.isEmpty else { return }
+    private func sendOutboundFrames(on path: QUICPath) {
```
tfpauly, line 3233: This function seems unnecessary now. Just call `path.sendEnqueuedOutboundDatagrams` directly.

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -3922,7 +3900,7 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
 
         QUICSignpost.outbound(id: signpostID, length: totalBytesWrittenInFrame)
 
-        outboundFrames.add(frame: outFrame)
+        enqueueOutboundFrame(frame: outFrame, path: path)
```
tfpauly: Let's make this path.enqueueOutboundFrame(outFrame)

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
tfpauly: Also should be `_ frame: ` so the name is elided

rnro (commented):
This looks like a great perf change, I think I agree with all of Tommy's comments 😄. The methods on an object which never access `self.` is a giveaway for these.

Sources/SwiftNetwork/QUIC/QUICConnection.swift
agnosticdev: Addressed in [fa6035c](https://github.com/apple/swift-network-evolution/pull/129/commits/fa6035ce982e0f7132f6675e78a5db6ebab601b2)

Sources/SwiftNetwork/QUIC/QUICConnection.swift
agnosticdev: Address in [fa6035c](https://github.com/apple/swift-network-evolution/pull/129/commits/fa6035ce982e0f7132f6675e78a5db6ebab601b2)

Sources/SwiftNetwork/Protocols/ProtocolDatagramHandlers.swift
agnosticdev: Added in [fa6035c](https://github.com/apple/swift-network-evolution/pull/129/commits/fa6035ce982e0f7132f6675e78a5db6ebab601b2)

Sources/SwiftNetwork/Protocols/ManyToManyProtocol.swift
agnosticdev: Added in [fa6035c](https://github.com/apple/swift-network-evolution/pull/129/commits/fa6035ce982e0f7132f6675e78a5db6ebab601b2)

Sources/SwiftNetwork/Protocols/FrameArray.swift
```diff
@@ -143,6 +143,14 @@ public struct FrameArray: ~Copyable {
         }
     }
 
+    // Make sure to keep the original capacity of the array to prevent reallocations
+    public mutating func drainArrayKeepingCapacity() -> FrameArray {
+        let count = self.count
+        let returnArray = self
+        self = FrameArray(capacity: count)
```
glbrntt, line 149: `count` is the number elements, not the underlying capacity, is using it for the capacity intentional? Should it be `self.frames.capacity` instead?

agnosticdev: It is intentionally using count, yes.  Check out the creation of `lowerSendQueue`, it now has a starting capacity on it too.  The intention here is that these FrameArray's should maintain a capacity based on how many frames were previously in lowerSendQueue.  That way the collection does not have to spend so many cycles reallocating to fit newly added frames.

glbrntt: So is the assumption that the `lowerSendQueue` sends approximately the same number of items each time? 

This doesn't _feel_ like it would be representative of real traffic patterns. and I'd be worried about one large or small batch negatively impacting perf because the next `FrameArray` is then badly sized. I wonder if even a simple heuristic like taking the median of the last three `FrameArray` sizes would be more effective?

glbrntt: Also: I think the comment on the function is misleading because the original _capacity_ isn't kept. Some capacity is kept but it might not be all of it.

agnosticdev: > So is the assumption that the `lowerSendQueue` sends approximately the same number of items each time?
> 
> This doesn't _feel_ like it would be representative of real traffic patterns. and I'd be worried about one large or small batch negatively impacting perf because the next `FrameArray` is then badly sized. I wonder if even a simple heuristic like taking the median of the last three `FrameArray` sizes would be more effective?

For larger transfers, yes, the send queue should be capped at 40 packets and if the congestion window is opened wide enough then we will fill this each time we send.  There can be error here, yes, but for larger transfers this methodology is effective, as seen in the data above.

Just to make sure I am not regressing performance for highly concurrent, multi-stream, traffic patterns I ran our QUICStreamLoad benchmark with this change and with top of tree, and the results are almost a wash.  So for these traffic patterns we would not be regressing performance.  
Top of tree:
```
1.25 G 49.2%	14.00 M	   specialized QUICConnection.buildSinglePacketForKeyState()	
1.72 G 16.4%	1.01 M	   specialized QUICConnection.sendApplicationFrames()	
4.00 M 100.0%	-	       drainArray
```

With this change
```
1.22 G 49.0%	14.00 M	   specialized QUICConnection.buildSinglePacketForKeyState()	
1.74 G 16.6%	-	       specialized QUICConnection.sendApplicationFrames()	
7.00 M 100.0%	-	       drainArrayKeepingCapacity
```

agnosticdev: > Also: I think the comment on the function is misleading because the original _capacity_ isn't kept. Some capacity is kept but it might not be all of it.

Removed comment in [e2963a9](https://github.com/apple/swift-network-evolution/pull/129/commits/e2963a92c5ebceae999dc78a8b9383453f4ddc76)

tfpauly (approved):

PR #131, opened 2026-09-05
SwiftQUIC: PERF: Avoid heap hit for storing FrameAckRanges

agnosticdev:
This PR introduces a new type called `NetworkSmallUniqueArray` so that when Ack frames are parsed we do not need to take a heap hit storing `FrameAckRanges`.  We can now store up to 5 ACK ranges before the collection fails over to using a dynamic array.  We see a decent CPU savings as well:
Top of tree:
```
109.43 M 100.0%	 - FrameAck	 (Total)

28.00 M 79.4%	-     	     FrameAck.init(frame:packetNumberSpace:)	
28.00 M 79.4%	2.00 M	       specialized FrameAck.init(frame:packetNumberSpace:)	
12.00 M 34.0%	-     	         specialized Array.init(repeating:count:)	
9.00 M  25.5%	2.00 M	         static Deserializer<>.deserialize<>(_:claim:_:)	
2.00 M  5.7%	-     	         protocol witness for static Equatable.== infix(_:_:) in conformance FrameType	
2.00 M  5.7%	2.00 M	         outlined init with copy of FrameAck	
1.00 M  2.8%	1.00 M	         outlined destroy of FrameAck	

```

With this change:
```
15.31 M 100.0%	-	 FrameAck (Total)	

8.00 M 78.3%	-     	    FrameAck.init(frame:packetNumberSpace:)	
7.00 M 68.5%	-     	     specialized FrameAck.init(frame:packetNumberSpace:)	
7.00 M 68.5%	1.00 M	     static Deserializer<>.deserialize<>(_:claim:_:)	
5.00 M 48.9%	-     	     static Deserializer<>.deserialize<>(_:_:)	
1.00 M 9.8%	1.00 M	         Frame.bytes.getter	
1.00 M 9.8%	1.00 M	         closure #2 in FrameAck.init(frame:packetNumberSpace:)	

2.00 M 100.0%	-	 NetworkSmallUniqueArray (Total)
```
Credit to @tfpauly for the NetworkSmallUniqueArray type.

### DIFF
--- a/Sources/SwiftNetwork/QUIC/Ack.swift
+++ b/Sources/SwiftNetwork/QUIC/Ack.swift
@@ -251,7 +251,7 @@ struct AckSpace: ~Copyable, PrefixedLoggable {
             if let ecnCounter {
                 ack.ecnCounter = ecnCounter
             }
-            ackFrame = ack
+            ackFrame = consume ack
         }
         var blockCount = 0
         for index in blocks.indices.reversed() {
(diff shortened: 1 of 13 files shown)


Sources/SwiftNetwork/QUIC/Ack.swift
```diff
@@ -711,40 +712,116 @@ struct Ack: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
...
         return shouldSend
     }
 
-    static func blockSequence(
-        frame: FrameAck,
-        oldestPacketNumber: PacketNumber = .initial
-    ) -> AckBlockSequence {
-        AckBlockSequence(
+    @inline(always)
+    private static func _forEachBlock<Ranges: AckRangeStorage & ~Copyable>(
```
rpaulo, line 730: It's already private static.  Why do we need `_` ?

Sources/SwiftNetwork/QUIC/Ack.swift
```diff
@@ -711,40 +712,116 @@ struct Ack: ~Copyable, PrefixedLoggable, NonCopyableTimerUser {
...
+            }
+            index += 1
+            if savedLargest >= oldestPacketNumber {
+                body((start: smallest, end: savedLargest))
+            }
+        }
+    }
+
+    /// Decodes ACK ranges directly using a `NetworkSmallUniqueArray<FrameAckRange, 3>.
+    /// Will avoid a heap allocation for ack range collection under 3
```
rpaulo, line 760: Why 3?

Sources/SwiftNetwork/QUIC/QUICFrame.swift
```diff
@@ -843,45 +828,43 @@ struct FrameAckRange {
     var range: PacketNumber
 }
 
+/// `FrameAckRange` is 16 bytes (two `PacketNumber`, each a wrapped `Int64`).
 @available(Network 0.1.0, *)
-struct FrameAck: QUICFrameProtocol {
-    var type = FrameType.ack
+extension FrameAckRange: NetworkInlineStorable {
+    typealias InlineSlot = InlineArray<2, UInt64>
```
rpaulo, line 834: I'm not sure how this helps since FrameAckRange is already 2 integers and nothing else?

Sources/SwiftNetwork/QUIC/Ack.swift
rpaulo: Also, this needs to be a type alias because we shouldn't need to write "3" (or whatever) everywhere.

Tests/QUICTests/AckTests.swift
```diff
@@ -433,7 +405,7 @@ final class AckTests: XCTestCase {
             delay: 0,
             ranges: [FrameAckRange(gap: 0, range: 0)]
         )
-        for block in Ack.blockSequence(frame: frame) {
+        Ack.forEachBlock(frame: frame) { block in
```
rpaulo, line 408: Would be better if we could do frame.forEachAckBlock {}

Tests/QUICTests/ECNTests.swift
```diff
@@ -400,6 +400,7 @@ extension ECNTestStepAck {
 }
 
 // MARK: ECN Validation Tests
+#if swift(>=6.4)
```
rpaulo, line 403: Why 6.4?

agnosticdev: On Swift 6.3.3 this runs into a compiler crash on Linux that is no present on Swift 6.4
The crash is:
```
/home/uadmin/swift-network-evolution-ack-ranges-part-one/Tests/QUICTests/ECNTests.swift:472:23: error: copy of noncopyable typed value. This is a compiler bug. Please file a bug with a small example of the bug
470 |         let newlyAckedECNCount = step.newlyAckedECNCount
471 |         switch step.frame {
472 |         case .ack(let ack):
    |                       `- error: copy of noncopyable typed value. This is a compiler bug. Please file a bug with a small example of the bug
473 |             for i in 0..<repeats {
```

agnosticdev:
Closing the PR and will split it into two pieces, the first with the non-copyable frame types, which does have some performance benefits.  The next will follow up when the fix for this makes its way to the tool chain: https://github.com/swiftlang/swift/pull/91664

PR #138, opened 2026-09-09
SwiftQUIC: switch from inline(__always) to inline(always)

agnosticdev:
This is finishing the conversion of switching to `inline(__always)` to `inline(always)`.


tfpauly (approved):

rpaulo:
Is inline(always) Swift >= 6.3 only?

agnosticdev:
> Is inline(always) Swift >= 6.3 only?

It does look like it landed in 6.3, yes.
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0496-inline-always.md

rpaulo (approved):

PR #141, opened 2026-09-10
SwiftQUIC: PERF: Optimize redundant code and exclusivity in the inbound path

agnosticdev:
There was several places in serviceReceivedDatagrams -> handleInbound -> handleInboundPacket where we can optimize the following:
1. Hoist variables up and pass them through the inbound path to avoid exclusivity
2. Avoid redundant stats gathering
3. Avoid finalizing a frame more than once
4. Avoid popFirst for each frame in the datagrams collection and wipe them all at once.

This consistently saves ~500 megacycles across multiples runs of QUICTransfer

Top of tree:
```
40.86 G 51.4%	-      	    protocol witness for ManyToManyOutboundDatagramProtocol.serviceReceivedDatagrams(path:) in conformance QUICConnection
40.85 G 51.4%	2.00 M 	      QUICConnection.serviceReceivedDatagrams(path:)
36.12 G 45.4%	-      	       specialized ManyToManyOutboundDatagramProtocol<>.accessReceivedDatagrams(path:_:)
36.11 G 45.4%	9.00 M 	        closure #1 in QUICConnection.serviceReceivedDatagrams(path:)
35.94 G 45.2%	81.15 M	         QUICConnection.handleInbound(frame:from:)
```

With this change
```
40.23 G 50.8%	-      	         protocol witness for ManyToManyOutboundDatagramProtocol.serviceReceivedDatagrams(path:) in conformance QUICConnection
40.22 G 50.8%	4.00 M 	           QUICConnection.serviceReceivedDatagrams(path:)
35.51 G 44.9%	-      	            specialized ManyToManyOutboundDatagramProtocol<>.accessReceivedDatagrams(path:_:)
35.49 G 44.9%	2.00 M 	             closure #1 in QUICConnection.serviceReceivedDatagrams(path:)
35.46 G 44.8%	-      	              FrameArray.iterateMutableFrames(_:)
35.46 G 44.8%	4.00 M 	               closure #1 in closure #1 in QUICConnection.serviceReceivedDatagrams(path:)
35.43 G 44.8%	30.00 M	                QUICConnection.handleInbound(frame:from:connectionState:isServerConnection:)
```


Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -1609,30 +1636,27 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
...
-            return
-        }
-
-        stats.increment(.rxBytes, by: dataLength)
         guard dataLength <= UInt16.max else {
             log.info("Refusing to parse packet with size \(dataLength)")
             return
         }
 
+        #if DatapathLogging
```
tfpauly, line 1642: This check seems odd, since log.datapath only does stuff if datapath logging is enabled already

Sources/SwiftNetwork/Protocols/FrameArray.swift
```diff
@@ -265,6 +265,11 @@ public struct FrameArray: ~Copyable {
         frames = .init()
     }
 
+    // NOTE: This function differs from finalizeAllFramesAsFailed because it assumes all frames are finalized already
+    public mutating func removeAll() {
```
tfpauly: I'm concerned about this — why not remove as we finalize? Leaving finalized frames in the array seems dangerous.

Sources/SwiftNetwork/QUIC/QUICConnection.swift
agnosticdev: This avoids an exclusivity check because log is a property of QUICConnection and this is on the hot path.  So it will not log, but it will check for exclusivity regardless.

Sources/SwiftNetwork/Protocols/FrameArray.swift
agnosticdev: I added this note in here to make sure callers were aware of the difference. In the use-case I have there cannot be frames that are not finalizes because `handleInbound` has a defer statement at the top of it that makes sure each frame becomes finalized:
```
defer {
    deferClosing = false
    frame.finalize(success: true)
}
```
So we should not run a non-finalized frame in this situation.  I would expect that most all situations would otherwise use `finalizeAllFramesAsFailed`.

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -1584,9 +1584,34 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
...
+            if connectionState.isTerminal {
+                log.debug(
+                    "Ignoring incoming packets for connection in terminal state"
+                )
+                datagrams.finalizeAllFramesAsFailed()
+                return
             }
+            var receivedBytes: Int = 0
+            let receivedPackets: Int = datagrams.count
+            datagrams.iterateMutableFrames { frame in
```
tfpauly, line 1599: There are two options that don't involve adding a `removeAll` footgun I'd like us to try:

1. Use the variant of `iterateMutableFrames` that returns `FrameIterationResult`, and for this loop, return `removeFrameAndContinue`. Please also change `frames.remove(at: index)` in `iterateMutableFrames` to special case `0`, and just use `removeFirst` for the `0` case (which is more efficient). This should avoid moving the frame on remove, because it never assigns it anywhere (just removes it)

If that isn't enough, we can try 2:

2. Create a new variant of `iterateMutableFrames`, called something like `iterateMutableFramesAndDispose`/`iterateMutableFramesAndRemoveAll` that loops through all of the frames (doesn't give you an option to stop looping) and removes all elements of the array at the end. This forces people to not call `removeAll` in isolation.

agnosticdev: Good instinct, using the special case `removeFirst` is roughly equivalent to using `removeAll`.
Here are the two options:
```
Remove First
24.00 M 0.0%	-      specialized UniqueDeque<>.removeFirst()	
7.00 M  0.0%	-      Frame.deinit	

iterateMutableFramesAndDestroy:
35.57 G 44.8%	-     	closure #1 in QUICConnection.serviceReceivedDatagrams(path:)	
35.57 G 44.8%	-     	  FrameArray.iterateMutableFramesAndDestroy(_:)	
35.53 G 44.8%	6.00 M	  closure #1 in closure #1 in QUICConnection.serviceReceivedDatagrams(path:)	
38.01 M 0.0%	-     	  specialized RigidDeque<>.deinit	
4.00 M  0.0%	-     	  specialized IndexingIterator.next()	
```
Switched to removeFirst in [ee6ab21](https://github.com/apple/swift-network-evolution/pull/141/commits/ee6ab216d1a88e8744f7b2e0d49698876b4f09c2)

Sources/SwiftNetwork/Protocols/FrameArray.swift
tfpauly: Please also add the same optimization here for the 0 index

agnosticdev: Addressed [5bda383](https://github.com/apple/swift-network-evolution/pull/141/commits/5bda383889f5b533d1f8c3dc09033a1153ef6ba3)

tfpauly (approved):

Sources/SwiftNetwork/Protocols/FrameArray.swift
```diff
@@ -116,11 +116,19 @@ public struct FrameArray: ~Copyable {
             case .stopIterating:
                 return
             case .removeFrameAndContinue:
-                frames.remove(at: index)
+                if index == 0 {
```
rpaulo: This optimization is not needed, see https://github.com/apple/swift-collections/blob/main/Sources/DequeModule/Deque/Deque%2BCollection.swift#L823

agnosticdev: Try it out, run a removeFirst and a removeAt(index:) 5 million times and record the CPU usage in Instruments and let me know what you see.

glbrntt: This optimization **_shouldn't_** be needed. If this code is meaningfully slower without it (and it sounds like it is) then upstreaming a patch to swift-collections might be a better approach than working around it here.

rpaulo: While we pursue the fix in swift-collections, let's hide this perf optimization in an extension:

```
extension Deque {
    @discardableResult
    @inline(always)
    mutating func fastRemove(at index: Int) -> Element {
        if index == 0 {
            return removeFirst()
        }
        return remove(at: index)
    }
}
```

agnosticdev: Addressed [ab55fc2](https://github.com/apple/swift-network-evolution/pull/141/commits/ab55fc279b5a0048181e36e53c55705680cf3f03)

rpaulo (approved):

PR #143, opened 2026-09-11
SwiftQUIC: Move FrameAck to non-copyable in preparation for non-copyable collection

agnosticdev:
This is a follow up to this PR: https://github.com/apple/swift-network-evolution/pull/131
This change adds the non-copyable benefits for `FrameAck` and `FrameStreamSendMetadata` and also lowers the memory layout of `FrameAck` and `SentPacketRecord`.

This is in preparation for adding a non-copyable collection to `FrameAck` to replace the heap allocated ranges.


Tests/QUICTests/ECNTests.swift
```diff
@@ -400,6 +400,7 @@ extension ECNTestStepAck {
 }
 
 // MARK: ECN Validation Tests
+#if swift(>=6.4)
```
agnosticdev, line 403: @rpaulo you had asked about this earlier, this is to fix a crash that appears on Linux (possibly Darwin too?) when running on Swift 6.3.3.
```
/home/../swift-network-evolution-ack-ranges-part-one/Tests/QUICTests/ECNTests.swift:472:23: error: copy of noncopyable typed value. This is a compiler bug. Please file a bug with a small example of the bug
470 |         let newlyAckedECNCount = step.newlyAckedECNCount
471 |         switch step.frame {
472 |         case .ack(let ack):
    |                       `- error: copy of noncopyable typed value. This is a compiler bug. Please file a bug with a small example of the bug
473 |             for i in 0..<repeats {
```

tfpauly (approved):

rpaulo (approved):

PR #146, opened 2026-09-14
Clock reads go via the scheduler

rnro:
This change threads clock reads through the scheduler and exposes new methods for reading the system time with names which clearly communicate that is what is happening.

This PR is probably best reviewed per-commit. The first shifts the reads to go through the scheduler, the second modifies `Pacer.getSendTime` to more efficiently use the new APIs and the third removes the process-global manual clock since matching control is now offered per-context.

### Motivation

Datapath code read `NetworkClock.Instant.now` directly, so no test could control/mock time. This makes tests slow and in many cases racey. 

### Description

The scheduler already owns the timers, so it is the natural owner of time as well.

  - `NetworkContext.Scheduler` gains now and `nowAbsolute`, with `DefaultScheduler` the one conformance allowed to read the real clock.
  - `NetworkClock.Instant.now` becomes package-visible `systemNow`, so a syscall read is more visible/intentional at the call site.
  - `QUICConnection` stamps the continuous and absolute clocks as a pair per batch, so `Pacer` converts between the domains without reading either - it previously read both once per paced packet.
  - The process-global manual clock behind `NETWORK_INTERNAL_TESTS` is removed. The scheduler gives the same control per context, so this replaces it rather than dropping it; `NETWORK_INTERNAL_TESTS` is gone from CI and the README.

### Performance

This change should be a marginal improvement but certainly not negative.

 - The read moves from a static that inlines within the module to a non-devirtualisable witness call, this should be indistinguishable.
 - Stamping the absolute clock alongside the continuous one on the batch removes two clock reads per-paced-packet.
 - This change causes no per-connection or per-packet allocation.

In test runs of `QUICTransfer` it seems to shave a few mega cycles, but hard to say with the spread:
```
Before:
    61.42G, 62.04G, 62.20G

After:
    61.51G, 61.85G, 61.31G
```


Tests/SwiftNetworkTests/SwiftNetworkClockTests.swift
```diff
@@ -327,123 +327,3 @@ final class SwiftNetworkClockTests: NetTestCase {
         }
     }
 }
-
```
rpaulo, line 330: Shouldn't we have a test that shows how we can manually advance the clock?

rpaulo:
Doesn't this make it more difficult to test smaller objects that don't have a NetworkContext to begin with? How would we test congestion control now?  Does that require us passing a NetworkContext to the congestion controller?

agnosticdev (approved):
Thank you for checking the performance results on this change!

rnro:
> Doesn't this make it more difficult to test smaller objects that don't have a NetworkContext to begin with? How would we test congestion control now? Does that require us passing a NetworkContext to the congestion controller?

I don't think so. We haven't had to add a context to parts of the codebase which have no interest in the context in this PR and the tests still pass for example. Because we have threaded `now` through the stack you can still use `NetworkClock.Instant.testBase`, `.advanced(by:))` to reference specific times without a context.

Tests/SwiftNetworkTests/SwiftNetworkClockTests.swift
rnro: Good point, thanks! This was split out of some larger work which exercised this code in tests but it should stand alone.

Sources/SwiftNetwork/Context/NetworkContext.swift
```diff
@@ -88,6 +88,17 @@ public final class NetworkContext: NetworkContextProtocol, @unchecked Sendable {
...
         var runningInScheduler: Bool { get }
+        /// The scheduler's current notion of the continuous clock.
+        ///
+        /// The scheduler owns time because it owns the timers: one that holds scheduled tasks
+        /// instead of arming an OS timer has to report the time it fires them at, or a deadline it
+        /// just ran would still look like it is in the future.
+        var now: NetworkClock.Instant { get }
+        /// The scheduler's current notion of the absolute clock.
+        ///
+        /// Separate from `now` because the two clocks diverge across system sleep, and `Pacer`
```
rpaulo: I would remove references to Pacer here since this is generic.  You could reference QUIC.

Sources/SwiftNetwork/Context/NetworkContext.swift
```diff
@@ -389,6 +400,13 @@ extension NetworkContext {
             // TODO: Not supported by DispatchQueue
             fatalError("Unsupported")
         }
+        /// The real clock. This is the one scheduler allowed to read it.
```
rpaulo: Sorry I'm having trouble understanding this sentence.  Can we explain this a bit more?

Sources/SwiftNetwork/QUIC/QLog.swift
```diff
@@ -734,8 +737,9 @@ final class QLog {
         bytesInFlight: UInt64 = UInt64.max,
         slowStartThresh: UInt64 = UInt64.max,
         packetsInFlight: UInt64 = UInt64.max,
-        timestamp: NetworkClock.Instant = .now
+        timestamp: NetworkClock.Instant? = nil
```
rpaulo, line 740: Do we need to pass it in qlog?  Why not always get it from the context?

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -252,19 +252,44 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
...
+    var nowAbsolute: NetworkClock.Instant {
+        pinnedClock?.absolute ?? context.nowAbsolute
+    }
+
+    /// Runs `body` with the clock pinned, so everything inside it sees one pair of readings.
+    ///
+    /// The outermost call takes the pin and releases it on the way out. One nested inside
+    /// another keeps the outer call's readings, so `now` holds still for the whole batch.
+    /// Code that needs real elapsed time within a batch reads `context.now` directly.
+    @inline(__always)
```
rpaulo: Didn't we discontinue `__always` ?

rpaulo (approved):
Approved pending comments.

Sources/SwiftNetwork/QUIC/QLog.swift
rnro: We pass in a timestamp so that the log matches the time at which the logged event actually happened, (or the batch it was in was processed) rather than the time we produced the log

Sources/SwiftNetwork/QUIC/QUICConnection.swift
rnro: Yep, good spot

Sources/SwiftNetwork/Context/NetworkContext.swift
rnro: Yep, I've fixed that mangled sentence.

tfpauly (approved):

PR #147, opened 2026-09-14
Fuzzing: add libFuzzer targets for QUIC packet and transport parameter parsing

rpaulo:
- Add `Fuzzing` package trait wiring fuzzer/ASan flags and two fuzz targets (`FuzzQUICPackets`, `FuzzTransportParameters`), plus CI and nightly workflow integration with corpus caching
- Add no-op `SecFramerNull` protector (FuzzBuild only) so fuzz inputs are treated as already-plaintext on the receive path
- Stop processing coalesced frames/packets once the connection is terminal, add `QUICFrame.discard`, guard missing TLS options on INITIAL retransmission, and clear crypto reassembly queues on stop


agnosticdev (approved):
Thanks!

Sources/SwiftNetwork/QUIC/Protector.swift
```diff
@@ -240,6 +243,56 @@ protocol SecFramerProtocol: ~Copyable {
     ) throws(QUICError)
 }
 
+#if Fuzzing
```
rnro, line 246: What do you think of this broken out into its own file `Protector+SecFramerNull.swift` or `Protector+Fuzzing.swift`

Sources/SwiftNetwork/QUIC/Protector.swift
```diff
@@ -157,9 +157,12 @@ enum TLSCipherSuite: CaseIterable {
 
 @available(Network 0.1.0, *)
 struct SecFramerKeys: ~Copyable {
-    enum KeyType {
+    enum KeyType: Equatable {
         case aesGCM
         case chaChaPoly
+        #if Fuzzing
```
rnro, line 163: This causes lots of `#if` statements throughout the code which I think in general makes a codebase hard to follow and reason about. I'm wondering if we could get away with unconditionally defining the `.null` case and having strategic `fatalError`s or similar which trigger if you use it in a non-fuzzing context. Maybe using an inlinable `assertFuzzing()` call with an `#if !Fuzzing` in it?

Sources/SwiftNetwork/QUIC/QUICConnection.swift
```diff
@@ -1712,7 +1712,15 @@ public final class QUICConnection: ManyToManyApplicationStreamProtocol,
                 break
             }
 
-            if closeError != nil {
+            if closeError != nil || state.isTerminal {
```
rnro, line 1715: Is this a bug which fuzzing surfaced? Are some of those included in this PR?

Sources/SwiftNetwork/QUIC/Protector.swift
rpaulo: We can do that.

Sources/SwiftNetwork/QUIC/Protector.swift
rpaulo: I really didn't want to compile it in !Fuzzing because it's somewhat dangerous to have this option.

Sources/SwiftNetwork/QUIC/QUICConnection.swift
rpaulo: Yes, I'm moving these fixes to https://github.com/apple/swift-network-evolution/pull/157#pullrequestreview-5239668467

Sources/SwiftNetwork/QUIC/Protector.swift
rnro: Can you expand on that? I'm not sure I follow.

rpaulo: I want the null fuzzer to never be compiled into production or tests.  Only when we are fuzzing.  I'm probably splitting hairs but not having the enum available outside of `Fuzzing` is a bit safer than fatalError().

PR #154, opened 2026-09-16
Frame: add hardware checksum offload and TSO metadata

rpaulo:
- Add RX transport checksum validation flags and hardware checksum value reporting
- Add TX partial-checksum offload start/store offsets
- Add TSO segment size and IPv6 super-packet support
- Fix inverted guard conditions in aggregate buffer length handling


Tests/SwiftNetworkTests/SwiftNetworkFrameTests.swift
```diff
@@ -18,7 +18,7 @@ import XCTest
 @available(anyAppleOS 27, *)
 final class SwiftNetworkFrameTests: XCTestCase {
     func testFrameLayout() {
-        XCTAssertEqual(MemoryLayout<Frame>.size, 136)
-        XCTAssertEqual(MemoryLayout<Frame>.stride, 136)
```
tfpauly, line 22: It would be great to find a way to not increase the size for most frames, but I don't see a great option at this point.

rpaulo: If this is a problem for embedded, maybe we could not support checksum offloading there and make this structure smaller?

rpaulo: I was able to move protocolMetadatas to UniqueArray which saved 8 bytes and converted bools to flags.

tfpauly (approved):

PR #156, opened 2026-09-17
SwiftQUIC: Adding QUICUtilities API for creating out of band version negotiation packet

agnosticdev:
Adds an API and test for a consumer of SwiftQUIC to create an out-of-band version negotiation packet so an entire QUICConnection does not need to be allocated.  This could be used in a server use-case.


Sources/SwiftNetwork/QUIC/QUICUtilities.swift
```diff
@@ -221,5 +221,44 @@ public struct QUICConnectionUtilities {
...
+    ///     - supportedVersions: The support version to advertise
+    public static func createVersionNegotiationPacket(
+        destinationConnectionID: QUICConnectionID,
+        sourceConnectionID: QUICConnectionID,
+        supportedVersions: [QUICVersion]
+    ) -> [UInt8] {
+        guard destinationConnectionID.length > 0, sourceConnectionID.length > 0, supportedVersions.count > 0 else {
+            Logger.proto.error("Failed to provide valid input, dcid, scid, or supported versions are empty")
+            return []
+        }
```
josephnoir: I think the guard is not correct. Both connection IDs can be 0 length per layout in [RFC 9000, Section 17.2.1](https://www.rfc-editor.org/rfc/rfc9000.html#section-17.2.1).

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
```diff
@@ -221,5 +221,44 @@ public struct QUICConnectionUtilities {
             return []
         }
     }
+
+    /// Creates a version negotiation packet from the given scid, dcid, and supported versions.
+    /// NOTE: Includes the negotiation pattern already.
```
josephnoir, line 226: While it is valid to include a reserved version pattern, it does not have to be (RFC 9000, [Section 6.3](https://www.rfc-editor.org/rfc/rfc9000.html#section-6.3-2), [Section 15](https://www.rfc-editor.org/rfc/rfc9000.html#section-15-7)). So, I am inclined to not have it included per default.

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
```diff
@@ -221,5 +221,44 @@ public struct QUICConnectionUtilities {
...
+    ///
+    /// - Parameters:
+    ///     - destinationConnectionID: The cid to include on the packet scid
+    ///     - sourceConnectionID: The cid to include on the packet dcid
+    ///     - supportedVersions: The support version to advertise
```
josephnoir:
```suggestion
    ///     - supportedVersions: The supported versions to advertise
```

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
agnosticdev: From a server's standpoint:
```
The server MUST include the value from the Source Connection ID field of the packet it receives
 in the Destination Connection ID field. The value for Source Connection ID MUST be copied from 
the Destination Connection ID of the received packet, which is initially randomly selected by a client. 
Echoing both connection IDs gives clients some assurance that the server received the packet and that the Version Negotiation packet was not generated by an entity that did not observe the Initial packet
```

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
agnosticdev: Thank you, addressed [1445baf](https://github.com/apple/swift-network-evolution/pull/156/commits/1445baf099c44b764202966a34b3a8da975dd028)

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
agnosticdev: I think we should keep the vn pattern here.

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
agnosticdev: I removed the guards, but just know that if you pass in a 0 length CID for these values you may not get a packet above 21 bytes which is the QUIC minimum length packet value.

josephnoir: Fair. After reading the RFC more [RFC 9000, Section 7.2](https://www.rfc-editor.org/rfc/rfc9000.html#section-7.2-3):

> When an Initial packet is sent by a client that has not previously received an Initial or Retry packet from the server, the client populates the Destination Connection ID field with an unpredictable value. This Destination Connection ID MUST be at least 8 bytes in length.

So, the `destinationConnectionID` cannot be zero-length in practice, even if the packet layout allows it? However, the SCID can be.

josephnoir (approved):
Thanks, @agnosticdev!

rpaulo:
I would argue that version negotiation is best handled by QUIC itself, not by an entity outside of QUIC because VN isn't stateless (see 9368).  I believe this mechanism, however, may be a good fit for RETRY packets.

rpaulo:
> I would argue that version negotiation is best handled by QUIC itself, not by an entity outside of QUIC because VN isn't stateless (see 9368). I believe this mechanism, however, may be a good fit for RETRY packets.

Sorry, I wasn't totally clear in this explanation: servers can handle incompatible versions in a stateless way but clients won't be able to.

josephnoir:
It's the server's side that I'd like this for. Ideally, swift-nio-quic can avoid setting up state to respond with a version negotiation. Would you like to make it clear that this is only intended for servers?

> I believe this mechanism, however, may be a good fit for RETRY packets.

That would be welcome!

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
```diff
@@ -221,5 +221,44 @@ public struct QUICConnectionUtilities {
             return []
         }
     }
+
+    /// Creates a version negotiation packet from the given scid, dcid, and supported versions.
```
rpaulo:
```suggestion
    /// Creates a version negotiation packet from the given SCID, DCID, and supported versions.
```

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
```diff
@@ -221,5 +221,44 @@ public struct QUICConnectionUtilities {
...
+    /// NOTE: Includes the negotiation pattern already.
+    ///
+    /// - Parameters:
+    ///     - destinationConnectionID: The cid to include on the packet scid
+    ///     - sourceConnectionID: The cid to include on the packet dcid
+    ///     - supportedVersions: The supported versions to advertise
+    public static func createVersionNegotiationPacket(
+        destinationConnectionID: QUICConnectionID,
+        sourceConnectionID: QUICConnectionID,
+        supportedVersions: [QUICVersion]
```
rpaulo: Shouldn't supportedVersions come from the QUIC stack?

agnosticdev: Addressed in [c4f060e](https://github.com/apple/swift-network-evolution/pull/156/commits/c4f060ea846a3726cbccda3cf730516e85ad4d5b)

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
agnosticdev: Addressed in [c4f060e](https://github.com/apple/swift-network-evolution/pull/156/commits/c4f060ea846a3726cbccda3cf730516e85ad4d5b)

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
```diff
@@ -221,5 +221,38 @@ public struct QUICConnectionUtilities {
             return []
         }
     }
+
+    /// Creates a version negotiation packet from the given SCID, DCID.
+    /// NOTE: Includes the negotiation pattern already and the supported versions of the QUIC stack.
+    ///
+    /// - Parameters:
+    ///     - destinationConnectionID: The cid to include on the packet scid
```
rpaulo, line 240: I think the API is a bit confusing because of DCID argument swapping.  Wouldn't it be better for the caller to do the swapping themselves? It would be more explicit that they are passing the right CID when they swapped it at the call site.  The API documentation should have a note about this.

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
```diff
@@ -221,5 +221,38 @@ public struct QUICConnectionUtilities {
...
+        // N.B.: The packet scid/dcid are swapped when constructing QUICVersionNegotiation
+        guard
+            let versionNegotiationPacket = try? QUICVersionNegotiation(
+                destinationConnectionID: sourceConnectionID,
+                sourceConnectionID: destinationConnectionID,
+                supportedVersions: versions
+            )
+        else {
+            Logger.proto.error("Failed to build a valid version negotiation packet")
+            return []
```
rpaulo: I know this is how the SR function is doing it, but this could be better handled with throw.

agnosticdev: Done in [1740fde](https://github.com/apple/swift-network-evolution/pull/156/commits/1740fdec71e88ed8f875b80dd423d6b86f765afa)

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
agnosticdev: Done in [1740fde](https://github.com/apple/swift-network-evolution/pull/156/commits/1740fdec71e88ed8f875b80dd423d6b86f765afa)

tfpauly (approved):

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
```diff
@@ -188,11 +188,15 @@ public struct QUICConnectionUtilities {
...
         triggeringPacketLength: Int
-    ) -> [UInt8] {
+    ) throws(NetworkError) -> [UInt8] {
 
         guard token.token.count > 0, triggeringPacketLength > 0 else {
-            Logger.proto.error("Failed to provide valid input: \(token), \(triggeringPacketLength)")
-            return []
+            throw NetworkError(
+                category: .init(
+                    identifier: "QUICUtilities.packetFailedToBuild",
```
rpaulo: Nit: I think just QUICUtilities would do.

rpaulo (approved):

Sources/SwiftNetwork/QUIC/QUICUtilities.swift
agnosticdev: Done: [bbb2b82](https://github.com/apple/swift-network-evolution/pull/156/commits/bbb2b823365fd4cb66754c9bb9dbfa1bd17602c1)

PR #157, opened 2026-09-17
QUIC: fix safety issues found by fuzzing

rpaulo:
- Stop processing coalesced packets and frames once the connection has reached a terminal state, avoiding use of torn-down per-connection state (e.g. crypto) after a graceful or error close
- Finalize embedded frames via new `QUICFrame.discard` when rejecting a frame before `processFrame`, preventing a fatal precondition
- Clear crypto reassembly queues on teardown and guard TLS restart on missing TLS options during INITIAL retransmit
- Gate transport parameter and stateless reset token error logging behind `DisableErrorLogging` for fuzzing builds


rpaulo:
Decided to separate the QUIC bug fixes from https://github.com/apple/swift-network-evolution/pull/147.

agnosticdev (approved):

tfpauly (approved):

PR #161, opened 2026-09-18
Requeue a datagram when it can't be sent

holleman:
Make sure to requeue a datagram and let the caller know that it could not be sent.  An example of this is when a pending ack has consumed space before the datagram.


Sources/SwiftNetwork/QUIC/SendItems.swift
```diff
@@ -1622,12 +1625,21 @@ extension FrameDatagram: SendableItem {
                             )
                         )
                     } catch {
-                        connection.log.error("Unable to write datagram for flow \(firstFlowID)")
+                        // Requeue rather than finalize and pass up the error so a retry is possible
+                        connection.log.datapath(
+                            "datagram for flow \(firstFlowID.debugDescription) does not fit \(roomBeforeWriting) free bytes; requeueing"
+                        )
+                        datagrams.prepend(frame: datagramFrame)
```
agnosticdev, line 1632: Okay, so here we just add the frame back in and frame array and that way pending items knows to build a packet here for that flow again, correct?

holleman: Yes, redo it.  If it was a pending ack it would have gone out and then when we return we can pick up an empty packet and then send the datagram.

agnosticdev (approved):
Thank you!

Sources/SwiftNetwork/QUIC/SendItems.swift
tfpauly: In general, this seems like a good idea to not drop locally -- but does this cause a case where not being able to send a particular datagram to be at the front of the list forever and if it is too big to fit, no other datagrams can send?

Sources/SwiftNetwork/QUIC/SendItems.swift
```diff
@@ -1622,12 +1625,21 @@ extension FrameDatagram: SendableItem {
                             )
                         )
                     } catch {
-                        connection.log.error("Unable to write datagram for flow \(firstFlowID)")
+                        // Requeue rather than finalize and pass up the error so a retry is possible
+                        connection.log.datapath(
+                            "datagram for flow \(firstFlowID.debugDescription) does not fit \(roomBeforeWriting) free bytes; requeueing"
```
rpaulo:
```suggestion
                            "Datagram for flow \(firstFlowID.debugDescription) does not fit \(roomBeforeWriting) free bytes; requeueing"
```

tfpauly (approved):

Sources/SwiftNetwork/QUIC/SendItems.swift
holleman: Done!

Sources/SwiftNetwork/QUIC/SendItems.swift
```diff
@@ -1622,12 +1625,21 @@ extension FrameDatagram: SendableItem {
                             )
                         )
                     } catch {
-                        connection.log.error("Unable to write datagram for flow \(firstFlowID)")
+                        // Requeue rather than finalize and pass up the error so a retry is possible
```
rpaulo, line 1628: The only reason why this fails is because we couldn't write, but since we already requested a big enough frame how can this fail?

